PR #3, opened 2017-06-02
Add ability to promote deployments

chdorner:
At the same time this introduces a new parameter called `TYPE` which
defines the type of deployment. This is so we can support the easiest
path for being able to promote QA deployments to prod.

This parameter can have three values:

* `promote` fetches the latest successful QA deployment and
  re-deploys it to QA, or promotes it to prod.
* `exact-version` deploys a specific version to the environmen.
* `sync-env` synchronizes the Elastic Beanstalk environment.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,6 +2,8 @@
 
 // The list of applications which can be deployed.
 def deployApplications = ['bouncer', 'via'].join('\n')
+// The list of deployment types.
+def deployTypes = ['promote', 'exact-version', 'sync-env'].join('\n')
 // The list of environments. It is assumed that each application has one of each
 // of these environments.
 def deployEnvironments = ['qa', 'prod'].join('\n')
@@ -13,12 +15,17 @@ pipeline {
         choice(name: 'APP',
                choices: deployApplications,
                description: 'Choose the application to deploy.')
+        choice(name: 'TYPE',
+               choices: deployTypes,
+               description: 'Choose the deployment type. ' +
+                            '`promote` promotes the last successful QA deployment to prod. ' +
+                            '`exact-version` pushes a specific docker tag. ' +
+                            '`sync-env` synchronizes the environment definition.',
+                defaultValue: 'promote')
         string(name: 'APP_DOCKER_VERSION',
                description: 'The tag of the application docker image to ' +
-                            'deploy. If you do not supply a tag, the ' +
-                            'environment configuration will be synchronised ' +
-                            'instead. (This can only happen if the ' +
-                            'environment has already been created.)',
+                            'deploy. This is required if the selected ' +
+                            'deployment type is `exact-version`.',
                defaultValue: '')
         choice(name: 'ENV',
                choices: deployEnvironments,
--- a/bin/eb-env-current-version
+++ b/bin/eb-env-current-version
@@ -1,0 +1,64 @@
+#!/usr/bin/env python
+
+"""
+Return the currently deployed docker tag in the given app and environment.
+"""
+
+import argparse
+import re
+import sys
+
+import boto3
+
+parser = argparse.ArgumentParser()
+parser.add_argument('app')
+parser.add_argument('env')
+
+
+def main():
+    args = parser.parse_args()
+
+    client = boto3.client('elasticbeanstalk')
+
+    app_version = fetch_application_version(client, args)
+    docker_tag = parse_docker_tag(app_version, args)
+    print(docker_tag)
+
+
+def fetch_application_version(client, args):
+    """Fetch current application version for specified environment."""
+    env_name = '{}-{}'.format(args.app, args.env)
+    result = client.describe_environments(ApplicationName=args.app,
+                                          EnvironmentNames=[env_name])
+    envs = result['Environments']
+
+    if len(envs) == 0:
+        abort('no matching environment found')
+
+    if len(envs) > 1:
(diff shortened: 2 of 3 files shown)


bin/jenkins
```diff
@@ -28,10 +34,28 @@ if [ -z "$ENV" ]; then
...
-    eb-deploy "$APP" "$ENV" "$APP_DOCKER_VERSION"
+if [ "$TYPE" = "exact-version" ]; then
+  if [ -z "$APP_DOCKER_VERSION" ]; then
+    abort "cannot proceed with exact-version deploy unless \$APP_DOCKER_VERSION is specified"
+  fi
+
+  eb-deploy "$APP" "$ENV" "$APP_DOCKER_VERSION"
+elif [ "$TYPE" = "promote" ]; then
+
+  app_version=$(aws elasticbeanstalk describe-environments --application-name $APP --environment-names $APP-qa |
```
chdorner: The initial goal would have been to store the docker tag at the end of the deploy build in case it was successful. A new promotion deployment would then retrieve this piece of data and deploy that docker version.

I couldn't find a way to programmatically store and retrieve data between builds in Jenkins (the Metadata plugin doesn't seem to have any API that I can use).

So the next best option is to query the current QA environment of the given application and parse out the docker tag (which we set [here](https://github.com/hypothesis/deployment/blob/f2e167bcad0c89d4911722809bc2b71759b22f31/bin/eb-deploy#L69)).

This way also has certain advantages, as it doesn't really matter what the state of Jenkins is. But rather it makes sure to get the version of the canonical place, which is Elastic Beanstalk.

I will probably pull this code out into a little helper and write that in Python, which would be much much nicer.

Jenkinsfile
```diff
@@ -13,12 +15,18 @@ pipeline {
         choice(name: 'APP',
                choices: deployApplications,
                description: 'Choose the application to deploy.')
+        choice(name: 'TYPE',
+               choices: deployTypes,
+               description: 'Choose the deployment type. `promote` re-deploys ' +
+                            'the current version for QA, or promotes the ' +
```
fatbusinessman: I get that having `promote` act as a redeployment in QA means the script gets to do the same in both environments, but it feels a little confusing to my brain. I can also see a case where someone would *think* they’ve promoted QA to production, whereas what they’ve actually done is redeploy QA.

As another option, could we say that `promote` builds will _only_ work when run against production, and then at some later point add a `redeploy` build type which will work against either environment?

bin/jenkins
```diff
@@ -28,10 +34,21 @@ if [ -z "$ENV" ]; then
...
+elif [ "$TYPE" = "promote" ]; then
+
+  docker_tag=$(eb-env-current-version "$APP" qa)
+  eb-deploy "$APP" "$ENV" "$docker_tag"
+
+elif [ "$TYPE" = "sync-env" ]; then
+  eb-env-sync "$APP" "$ENV"
 else
-    eb-env-sync "$APP" "$ENV"
+  abort "Don't know how to handle deployment type \"$TYPE\""
```
fatbusinessman: Fail-safe defaults FTW!

bin/eb-env-current-version
```diff
@@ -0,0 +1,64 @@
...
+
+    if len(envs) > 1:
+        abort('multiple environments found, but this tool only works '
+              'if environment names are unique within an application')
+
+    return envs[0]['VersionLabel']
+
+
+def parse_docker_tag(app_version, args):
+    exp = '^%s-[0-9]{8}T[0-9]{6}Z-(.*)$' % args.app
```
chdorner: We're setting the version label [here](https://github.com/hypothesis/deployment/blob/master/bin/eb-deploy#L69). But good point about the comment, I have added one.

chdorner: :+1:

bin/eb-env-current-version
```diff
@@ -0,0 +1,64 @@
...
+    result = re.search(exp, app_version)
+    if result is None:
+        abort('Found an incorrect application version "%s"' % app_version)
+
+    docker_tag = result.group(1)
+    if len(docker_tag) == 0:
```
chdorner: Good point. Looks much nicer like this :+1:

bin/jenkins
```diff
@@ -28,10 +34,21 @@ if [ -z "$ENV" ]; then
...
-    eb-deploy "$APP" "$ENV" "$APP_DOCKER_VERSION"
+if [ "$TYPE" = "exact-version" ]; then
+  if [ -z "$APP_DOCKER_VERSION" ]; then
+    abort "cannot proceed with exact-version deploy unless \$APP_DOCKER_VERSION is specified"
+  fi
+
+  eb-deploy "$APP" "$ENV" "$APP_DOCKER_VERSION"
+elif [ "$TYPE" = "promote" ]; then
+
+  docker_tag=$(eb-env-current-version "$APP" qa)
```
chdorner: Yes, because of `set -eu`, specifically because of the `-e`. If any command fails in this script, then the script will immediately exit.

Jenkinsfile
chdorner: Sure. That's done as well.

bin/jenkins
fatbusinessman: > Yes, because of `set -eu`

Ah yes—got it. Somehow I completely missed that line when I looked through the file before; I think I was looking for it right at the top.

PR #4, opened 2017-06-08
Notify Slack when deploying

chdorner:
This is using the `slackSend` step that is provided by the [Jenkins
Slack plugin](https://github.com/jenkinsci/slack-plugin).

First, this sends a start message and then depending on the state of the
job either a success or error message to the Slack channel that is
defined in the Jenkins integration in Slack itself.

You can check example messages in the [#jenkins-test](https://hypothes-is.slack.com/archives/C5P8WLLRE/p1496923995303213) channel.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,6 +8,22 @@ def deployTypes = ['promote', 'exact-version', 'sync-env'].join('\n')
 // of these environments.
 def deployEnvironments = ['qa', 'prod'].join('\n')
 
+def postSlack(state, params) {
+    def messages = [
+        'start': ['promote': "Starting to promote ${params.APP} to ${params.ENV}",
+                  'exact-version': "Starting to deploy ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV}",
+                  'sync-env': "Starting to synchronize the ${params.APP}-${params.ENV} environment"],
+        'success': ['promote': "Successfully promoted ${params.APP} to ${params.ENV}",
+                  'exact-version': "Successfully deployed ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV}",
+                  'sync-env': "Successfully synchronized the ${params.APP}-${params.ENV} environment"],
+        'error': ['promote': "Failed to promote ${params.APP} to ${params.ENV}",
+                  'exact-version': "Failed to deploy ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV}",
+                  'sync-env': "Failed to synchronize the ${params.APP}-${params.ENV} environment"]
+    ]
+    def colors = ['start': 'good', 'success': 'good', 'error': 'danger']
+    slackSend color: colors[state], message: messages[state][params.TYPE]
+}
+
 pipeline {
     agent { dockerfile true }
 
@@ -36,6 +52,12 @@ pipeline {
     }
 
     stages {
+        stage('setup') {
+            steps {
+                postSlack('start', params)
+            }
+        }
+
         stage('main') {
             steps {
                 withCredentials([[$class: 'AmazonWebServicesCredentialsBinding',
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -8,6 +8,22 @@ def deployTypes = ['promote', 'exact-version', 'sync-env'].join('\n')
 // of these environments.
 def deployEnvironments = ['qa', 'prod'].join('\n')
 
+def postSlack(state, params) {
+    def messages = [
+        'start': ['promote': "Starting to promote ${params.APP} to ${params.ENV}",
+                  'exact-version': "Starting to deploy ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV}",
+                  'sync-env': "Starting to synchronize the ${params.APP}-${params.ENV} environment"],
+        'success': ['promote': "Successfully promoted ${params.APP} to ${params.ENV}",
```
fatbusinessman: Is there any way to get the promotion success message to say which version got promoted? I realise this isn’t available in the parameters, but is there any way to get hold of the `docker_tag` we’ve retrieved from QA?

Jenkinsfile
```diff
@@ -8,6 +8,22 @@ def deployTypes = ['promote', 'exact-version', 'sync-env'].join('\n')
 // of these environments.
 def deployEnvironments = ['qa', 'prod'].join('\n')
 
+def postSlack(state, params) {
+    def messages = [
+        'start': ['promote': "Starting to promote ${params.APP} to ${params.ENV}",
+                  'exact-version': "Starting to deploy ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV}",
```
fatbusinessman: If a parameter is missing, the message currently reads, for example, “Starting to deploy bouncer missing to qa”. While this makes sense once I think about it, it’s a little confusing at first glance. Do you know whether there’s a way to have it use something other than “missing” for the default value without over-complicating the configuration?

Jenkinsfile
```diff
@@ -8,6 +8,22 @@ def deployTypes = ['promote', 'exact-version', 'sync-env'].join('\n')
...
+                  'sync-env': "Starting to synchronize the ${params.APP}-${params.ENV} environment"],
+        'success': ['promote': "Successfully promoted ${params.APP} to ${params.ENV}",
+                  'exact-version': "Successfully deployed ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV}",
+                  'sync-env': "Successfully synchronized the ${params.APP}-${params.ENV} environment"],
```
fatbusinessman: Is this naming convention (`bouncer-qa`, for example) the convention we’re using for environments in Elastic Beanstalk? If that’s the case, cool.

Jenkinsfile
chdorner: Ah, sorry for the confusion. I deliberately typed `missing` as the docker version. To make the deploy fail so I could test the error message.

Jenkinsfile
chdorner: Not without rewriting the way we promote. I was thinking of this as well. The only way I can see this is that if the type is `promote` then we fetch the current tag inside of the Jenkins file and then have the `exact-version` and `promote` basically do the same thing.

I'd suggest we keep this in mind for a later point? What do you think?

Jenkinsfile
chdorner: Yep, it's set done [here](https://github.com/hypothesis/deployment/blob/master/bin/eb-env-create#L68).

Jenkinsfile
fatbusinessman: > I'd suggest we keep this in mind for a later point? What do you think?

Yup: agreed. It would be a nice enhancement in the future, but for the time being I would think we’ll be absolutely fine looking up through the log to find the most recent `qa` version. If that turns out not to be fine, we can prioritise looking into ways to do the enhanced version.

PR #6, opened 2017-06-12
Bootstrap h-periodic qa and prod environments

chdorner:


h-periodic/env-prod.yml
```diff
@@ -0,0 +1,37 @@
...
+  aws:elasticbeanstalk:environment:
+    EnvironmentType: LoadBalanced
+    LoadBalancerType: application
+    ServiceRole: aws-elasticbeanstalk-service-role
+  aws:elasticbeanstalk:environment:process:default:
+    HealthCheckPath: /_status
+  aws:elasticbeanstalk:healthreporting:system:
+    SystemType: enhanced
+  aws:ec2:vpc:
+    Subnets: subnet-0f19e456,subnet-ee2c418b
```
chdorner: These are our private subnets

h-periodic/env-prod.yml
```diff
@@ -0,0 +1,37 @@
...
+  aws:elasticbeanstalk:environment:
+    EnvironmentType: LoadBalanced
+    LoadBalancerType: application
+    ServiceRole: aws-elasticbeanstalk-service-role
+  aws:elasticbeanstalk:environment:process:default:
+    HealthCheckPath: /_status
+  aws:elasticbeanstalk:healthreporting:system:
+    SystemType: enhanced
+  aws:ec2:vpc:
+    Subnets: subnet-66b8413f,subnet-2fdcbe4a
```
chdorner: These are our private subnets + I've used the same for the ELB subnets

h-periodic/env-prod.yml
```diff
@@ -0,0 +1,37 @@
...
+    ListenerEnabled: True
+    DefaultProcess: default
+  aws:autoscaling:launchconfiguration:
+    SecurityGroups: sg-fdf9c19a
+    IamInstanceProfile: aws-elasticbeanstalk-ec2-role
+    InstanceType: t2.nano
+    EC2KeyName: ops
+  aws:autoscaling:asg:
+    MaxSize: '1'
+    MinSize: '1'
```
chdorner: We lock the autoscaling to only ever have one instance running.

h-periodic/env-prod.yml
nickstenning: In an ideal world we'd bring this up as a [single-instance](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/using-features-managing-env-types.html#single-instance-environ) environment, to avoid paying the $30/month for the load balancer. Is it worth looking into that now, do you think?

chdorner: I did look into that, the issue with single-instance is that you can't define a health check. It just uses the two EC2 instance checks. The health check makes sure that we can connect to RabbitMQ.

nickstenning: Guh.

nickstenning: Thanks, Amazon.

nickstenning:
LGTM. Once we've migrated this stuff over I suspect it's time to have a look into getting rid of the need to run this application.

PR #8, opened 2017-06-22
Split eb-release out of eb-deploy

nickstenning:
Deploying an application to an EB environment consists of two main steps:

1. Creating an "application version" in Elastic Beanstalk corresponding to a particular Docker image tag.
2. Triggering an environment update in which the environment is configured to use the application version created in the previous step.

This PR moves the code associate with part 1 out of `eb-deploy` and into a separate script, `eb-release`, which returns the created version label on STDOUT.

This simplifies implementation of the "promote" deployment type and will simplify implementation of a "redeploy" deployment type, as these can then deal in the version labels used by Elastic Beanstalk natively, rather than constantly having to translate back and forth between Docker tag and version label.

### DIFF
--- a/bin/eb-release
+++ b/bin/eb-release
@@ -1,0 +1,85 @@
+#!/bin/sh
+
+usage () {
+    echo "Usage: $(basename "$0") <APP> <APP_DOCKER_TAG>"
+    echo
+    echo "Create an application version for the named application using a"
+    echo "specified Docker image tag."
+    echo
+    echo "Returns the app version created in Elastic Beanstalk."
+    echo
+    echo "  APP              the name of the application, e.g. 'bouncer'"
+    echo "  APP_DOCKER_TAG   the tag of the Docker image to release"
+}
+
+abort () {
+    echo "Error:" "$@" >&2
+    echo "Aborting!" >&2
+    exit 1
+}
+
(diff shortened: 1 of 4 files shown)


bin/eb-release
```diff
@@ -0,0 +1,85 @@
...
+status () {
+    echo "--->" "$@" >&2
+}
+
+if [ "$#" -ne 2 ]; then
+    usage >&2
+    exit 1
+fi
+
+set -eu
```
robertknight: Any particular reason `set -eu` is here rather than at the top of the script?

robertknight:
I haven't actually tested this script out, but I understand what's going on and the changes look fine to me. I leave it as an exercise for the PR author to blow stuff up in production.

bin/eb-release
nickstenning: Good question. In this specific case there isn't any good reason, no. In other scripts it's because variable defaults get declared higher up the file (e.g. [here](https://github.com/hypothesis/deployment/blob/74349f90b2ec1ea13606d030076fb66745035bb4/bin/jenkins#L3-L17)) and the `set -eu` *has* to come after that otherwise those lines will bomb out rather than falling through to defaults.

PR #9, opened 2017-06-23
Replace 'promote' deploy type with 'redeploy'

nickstenning:
The way that the deployment pipelines for individual applications now work means that a separate "promote" deployment type isn't that useful.

What is very useful, however, is a deployment type that simply triggers a redeployment of the currently deployed version to an environment, thus cycling the infrastructure for that environment.

This commit:

- Removes the 'promote' deployment type
- Adds a 'redeploy' deployment type
- Renames the 'exact-version' deployment type to simply 'deploy' so it's clearer what is going on in Jenkins logs.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,21 +3,21 @@
 // The list of applications which can be deployed.
 def deployApplications = ['bouncer', 'h', 'h-periodic', 'metabase', 'via'].join('\n')
 // The list of deployment types.
-def deployTypes = ['promote', 'exact-version', 'sync-env'].join('\n')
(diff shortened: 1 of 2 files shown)


fatbusinessman:
> The way that the deployment pipelines for individual applications now work means that a separate "promote" deployment type isn't that useful.

What’s changed? Is there an explanation somewhere of what the new pipelines look like?

nickstenning:
Hopefully the demo just now answers this question? We previously thought we'd do automated deployments to QA in the app build pipeline, and then an entirely separate "promote" step in a different job.

As it turns out, we can elegantly handle QA deployments, then waiting for user input, then deploying to prod, all from an application build.

PR #11, opened 2017-06-29
Ability to run remote commands

chdorner:
These commits install the Amazon SSM agent on the Elastic Beanstalk instances and add an `eb-run-task` script which can be used to execute remote commands.

The infrastructure changes that are needed (new policy to IAM instance role, S3 bucket) are done in hypothesis/playbook#142

### DIFF
--- a/common/ebextensions/tasks.config
+++ b/common/ebextensions/tasks.config
@@ -1,0 +1,3 @@
+packages:
+  rpm:
+    amazon-ssm-agent: https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/linux_386/amazon-ssm-agent.rpm
--- a/common/ebextensions/tasks.config
+++ b/common/ebextensions/tasks.config
@@ -1,3 +1,39 @@
 packages:
   rpm:
     amazon-ssm-agent: https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/linux_386/amazon-ssm-agent.rpm
+files:
+  "/usr/local/sbin/run-docker-task":
+    mode: 000770
+    owner: root
+    group: root
+    content: |
+      #!/bin/sh
+
+      set -eu
+
+      usage () {
+          echo "Usage: $(basename "$0") <COMMAND>"
+          echo
+          echo "Run a command inside the first running Docker container."
+          echo
+          echo "  COMMAND        the command to run, e.g. 'hypothesis user admin myuser'"
+          echo
+      }
+
+      if [ "$#" -lt 1 ]; then
+          usage >&2
+          exit 1
+      fi
+
+
+      # Fetch id of first running container
+      CONTAINER_ID=$(docker ps --filter "status=running"  --format "{{.ID}}" | head -n1)
+
+      if [ -z "$CONTAINER_ID" ]; then
+          echo "Error: Cannot find a running container" >&2
+          exit 1
+      fi
+
+
+      # Execute command
+      docker exec "$CONTAINER_ID" "$@"
--- a/bin/eb-release
+++ b/bin/eb-release
@@ -53,13 +53,26 @@ fi
 
 status "preparing application source bundle"
 
+SRCBUNDLEDIR=$(mktemp -d)
+
 # Use JQ to append ":<APP_DOCKER_TAG>" to the image name in the template JSON
 # file.
-SRCBUNDLE=$(mktemp)
-jq ".Image.Name += \":${APP_DOCKER_TAG}\"" <"${APP}/Dockerrun.aws.json" >"$SRCBUNDLE"
+jq ".Image.Name += \":${APP_DOCKER_TAG}\"" <"${APP}/Dockerrun.aws.json" >"$SRCBUNDLEDIR/Dockerrun.aws.json"
+
+# Copy additional configuration
+EBEXTENSIONS="${SRCBUNDLEDIR}/.ebextensions/"
+if [ -d "common/ebextensions" ]; then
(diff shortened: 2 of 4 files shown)


bin/eb-run-task
```diff
@@ -0,0 +1,96 @@
+#!/bin/sh
```
nickstenning: Just a small thing, but maybe calling this `eb-task-run` would be a little nicer from a command grouping perspective (i.e. it would match `eb-task-wait`).

bin/eb-run-task
```diff
@@ -0,0 +1,96 @@
...
+
```
nickstenning: I think you probably want a `set -eu` in here somewhere.

bin/eb-run-task
```diff
@@ -0,0 +1,96 @@
...
+  abort "Could not find a running instance for $APP and $ENV"
+fi
+
+
+# Protect instance from a scale-in event
+
+status "protecting instance from a scale-in event"
+
+ASG_NAME=$(aws elasticbeanstalk describe-environment-resources \
+               --environment-name ${APP}-${ENV} \
```
nickstenning: Probably worth quoting `"${APP}-${ENV}"` here.

bin/eb-run-task
```diff
@@ -0,0 +1,96 @@
...
+               --query "EnvironmentResources.AutoScalingGroups[0].Name" \
+               --output text)
+
+aws autoscaling set-instance-protection \
+    --auto-scaling-group-name "${ASG_NAME}" \
```
nickstenning: No need to change this at all, but my personal preference would be to omit the braces when they're not needed (such as when a variable is on its own in quotes).

bin/eb-run-task
```diff
@@ -0,0 +1,96 @@
...
+    --instance-ids "${INSTANCE_ID}" \
+    --protected-from-scale-in
+
+trap 'status "removing instance protection"; \
+      aws autoscaling set-instance-protection \
```
nickstenning: I can't actually think of any reason not to do this, but something about using a `trap ... EXIT` for this seems a bit weird to me. My inclination would be to move this to an explicit step at the end, but that's just me. I can't think of a solid reason why not to do this.

common/ebextensions/tasks.config
```diff
@@ -0,0 +1,39 @@
...
+files:
+  "/usr/local/sbin/run-docker-task":
+    mode: 000770
+    owner: root
+    group: root
+    content: |
+      #!/bin/sh
+
+      usage () {
+          echo "Usage: $(basename "$0") <COMMAND>"
```
nickstenning: The usage message seems a bit at odds with how this is actually used. Specifically, to me

```
Usage: $(basename "$0") <COMMAND>
```

and particularly

```
COMMAND        the command to run, e.g. 'hypothesis user admin myuser'
```

would imply that the command is a single argument (i.e. single-quoted string). But later you do:

```
COMMAND=$*
```

So if I'm reading this correctly you're actually expecting this to be invoked as, for example:

```
run-docker-task hypothesis user admin myuser
```

The documentation mismatch is only a little thing. I think the bigger thing to worry about is that you could run into some pretty weird quoting issues here. Here's an example program which behaves the same way as this one and shows how the shell interprets quotes:

```sh
#!/bin/sh

COMMAND=$*

set -x
: docker exec containerid sh -c "$COMMAND"
```

This works just fine if none of the args to the command are quoted:

```
$ ./quotetest foo bar baz
+ : docker exec containerid sh -c 'foo bar baz'
```

but you rapidly run into weirdness if you try and pass quoted args:

```
$ ./quotetest foo bar 'quoted string'
+ : docker exec containerid sh -c 'foo bar quoted string'
```

(That's three arguments to `quotetest` that are going to be treated by `sh -c` as four.)

I reckon the easiest way to avoid this is to ensure you use `$@` rather than `$*`, and you can only really do this if you don't store the command in an intermediate variable. In addition, passing args to `sh -c` gets tricky, so I'd probably skip the subshell altogether:

```
#!/bin/sh

set -x
: docker exec containerid "$@"
```

That then works as you'd hope with quoted args:

```
$ ./quotetest foo bar 'quoted string'
+ : docker exec containerid foo bar 'quoted string'
```

bin/eb-run-task
chdorner: I did this in a `trap ... EXIT` mainly because it is quite important to ensure that the instance protection gets removed. Otherwise, this instance would never get replaced until the next deployment. Maybe "quite important" is a bit of an overstatement. If we leave the instance protection in place a scale-in event would not remove the oldest instance, but the oldest that is not protected, which under some circumstances could mean that we have instances running for a very long time (which is not a problem in itself).

nickstenning: Ah, sorry, I see. So you're ensuring that it runs if the program executes early due to `set -eu`. I just had to write this to prove to myself that's what `trap ... EXIT` does:

```sh
#!/bin/sh

set -eu
trap 'echo trap called' EXIT
test 1 -eq 0
```

Perhaps in that case it might at least be worth putting the cleanup into a named function?

```sh
#!/bin/sh

set -eu

cleanup() {
    echo trap called
}

trap cleanup EXIT

test 1 -eq 0
```

chdorner: Yup, I've put it in a specific clean up function now.

common/ebextensions/tasks.config
chdorner: Yes, that's much better! I think at some point I gave up trying to get all the different quoted strings to work.

PR #13, opened 2017-07-04
Add Jenkins pipeline for running tasks on h instances

chdorner:
This is the Jenkins portion of the work that started in #11.

We can then configure a Jenkins Pipeline project that uses this script
which has a couple of common tasks pre-configured that we can run
against either the qa or prod environment.

For testing purposes, I already created the Jenkins project which is using this branch and is available at https://jenkins.hypothes.is/job/h-tasks/
Note: Until a new h version is deployed to production the tasks will only work in qa, as prod instances don't have the SSM agent installed yet.

### DIFF
--- a/h/Jenkinsfile-tasks
+++ b/h/Jenkinsfile-tasks
@@ -1,0 +1,89 @@
+#!groovy
+
+def environments = ['qa', 'prod'].join('\n')
+
+def tasks = [
+    'db current revision (timeout 10m)': [cmd: 'migrate current', timeout: '600'],
+    'db upgrade to head (timeout 10m)': [cmd: 'migrate upgrade head', timeout: '600', confirm: true],
+    'db upgrade to next (timeout 10m)': [cmd: 'migrate upgrade +1', timeout: '600', confirm: true],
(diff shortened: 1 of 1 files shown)


h/Jenkinsfile-tasks
```diff
@@ -0,0 +1,89 @@
+#!groovy
+
+def environments = ['qa', 'prod'].join('\n')
+
+def tasks = [
+    'db current revision (timeout 10m)': [cmd: 'migrate current', timeout: '600'],
+    'db upgrade to head (timeout 10m)': [cmd: 'migrate upgrade head', timeout: '600', confirm: true],
+    'db upgrade to next (timeout 10m)': [cmd: 'migrate upgrade +1', timeout: '600', confirm: true],
+    'db downgrade to previous (timeout 10m)': [cmd: 'migrate downgrade -1', timeout: '600', confirm: true],
+    'search reindex (timeout 2h)': [cmd: 'migrate downgrade -1', timeout: '7200', confirm: true]
```
nickstenning: That could be a surprising side-effect of running a search reindex... 😛

chdorner: Uhm, yes. Maybe we should avoid this surprising side-effect. Fixed.

PR #17, opened 2017-12-21
Add second target channel for Jenkins messages

segdeha:
While we have Atomic Jolt engaged with us working on the LTI app, it would be helpful to have Jenkins builds report into the channel to which they have access (#feat-canvas).

This change adds a second call to `slackSend` with an additional parameter, `channel`, set to the name of the channel. In theory, this should send all Jenkins messages for the `lti` repo to both #eng-general and #feat-canvas.


seanh:
@segdeha Didn't you now change it so that it reports deployments but not management tasks? Gonna merge this anyway, so we'll see.

segdeha:
Yes, the most recent commit has me moving the change to the root Jenkinsfile, which should report deployments rather than management tasks.

PR #30, opened 2018-12-11
Add Sakai to set of deployable apps

robertknight:
Add config to deploy Sakai image built from https://github.com/hypothesis/sakai-docker.

This will give us a persistent version of Sakai for testing the LMS app which is
available to everyone at Hypothesis.

We don't really need a QA environment for Sakai given that it is only
used for internal testing, but the Jenkinsfile script currently assumes
that both exist. We could save a few dollars by removing that
limitation.


sakai/env-prod.yml
```diff
@@ -0,0 +1,45 @@
+AWSConfigurationTemplateVersion: 1.1.0.0
```
robertknight: This configuration file was copied from metabase. It's fine for the most part, but I need to check whether the subnets are metabase-specific or not.

robertknight: If the subnets are shared with other EB apps we'll probably need to set up separate subnets for this deployment so that Sakai can't access Elasticsearch or Postgres. Postgres has user + password auth as well as security group traffic controls in the VPC, but it would be better to have appropriate security group controls in place.

robertknight:
The security groups referenced in the env-{prod, qa} files are currently the metabase ones. They will need to be updated to reference the ones created by https://github.com/hypothesis/playbook/pull/187

robertknight:
> The security groups referenced in the env-{prod, qa} files are currently the metabase ones.

This is resolved.

robertknight:
We'll also need a MySQL DB for this to work. My plan is to set up the smallest possible RDS MySQL instance for this.

robertknight:
I'll un-"Do not merge" this once the MySQL DB is set up.

robertknight:
Since this is not a user-facing service, I should also change the deployment policy to whatever will deploy fastest, rather than the "Immutable" default we've been using for other projects.

PR #31, opened 2018-12-12
Add dedicated security groups for bouncer app instances

robertknight:
This will enable giving Bouncer access to resources (eg. Elasticsearch)
without giving other EB services access to them.

See https://hypothes-is.slack.com/archives/C4K6M7P5E/p1544623705096100


robertknight:
See also https://github.com/hypothesis/playbook/pull/189. Discussed [on Slack](https://hypothes-is.slack.com/archives/C4K6M7P5E/p1544623705096100).

PR #33, opened 2019-04-16
add papertrail log forwarding service

dmfine:
If applied, this will add steps to the deployment of ALL APPS which use this pipeline.

The .ebextensions run on each container at startup will setup a services which forwards select logfiles to papertrail in realtime.


dmfine:
> This looks good to me. I'm a little concerned about applying it to all apps at first though. Is it possible to apply it to just one of the apps, say lms, first and then if it works apply it to the others?

I agree. I created a new PR with changes to make that possibe.

PR #36, opened 2019-04-19
fix: rename config file extension

dmfine:


dmfine:
just gonna merge a fix for incorrect file extension that wasn't caught in previous review

PR #39, opened 2019-05-17
Fix requirements handling

seanh:
Use a requirements.in file and compile it to requirements.txt.

Also pin PyYAML to 3.13. This is necessary to avoid a version conflict with awscli.


requirements.txt
```diff
@@ -1,3 +1,20 @@
...
+docutils==0.14
+futures==3.2.0
+jmespath==0.9.4
+pyasn1==0.4.5
+python-dateutil==2.8.0
+pyyaml==3.13
+rsa==3.4.2
+s3transfer==0.2.0
+six==1.12.0
+urllib3==1.24.3
```
robertknight: Hmm - Didn't pip-compile used to list the direct dependencies that caused indirect dependencies not explicitly listed in requirements.in to be included?

seanh: Hmm re-ran with `--annotate` to add them

PR #43, opened 2019-06-28
Add newrelic nginx integration

dmfine:
If committed, this ebextension configuraton will install and configure the NewRelic Nginx integration
which sends metrics gathered from the /status endpoint.

### DIFF
--- a/h/ebextensions/qa/nri_nginx.conf
+++ b/h/ebextensions/qa/nri_nginx.conf
@@ -57,10 +57,6 @@ files:
         }
 
 commands:
-# create infra config file
-  "00-config-file"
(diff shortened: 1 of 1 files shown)


robertknight:
Hi @dmfine - Is there someone specifically on-point to review this?

hmstepanek:
The data we are getting right now is not really helpful so I'm thinking we should just forgo adding this and explore other options. I'd really like to have something that reports information specifically on nginx rate limiting information; aka how many requests are in the queue, how many requests have been sent, how many requests were rejected, how full the queues are, etc.

dmfine:
Looks like it's not that useful, so closing

PR #44, opened 2019-09-04
Increase lms-qa EC2 instance size

robertknight:
lms-qa deployments were taking an unexpectedly long time (5-6 minutes).
Observing a deployment indicated this might be due to the instance
running out of memory when two containers (current and new deployment)
were running at once.

Once the new deploy is up and running, everything seemed fine
performance/memory-usage wise.

Attempt to resolve this by increasing the EC2 instance size to one with double the memory (1GB instead of 512MB)


robertknight:
FWIW most of our other apps are already using larger instance sizes (h, via).

PR #45, opened 2019-09-12
Switch lms-qa back to rolling deployments

robertknight:
Speed up our deployment process for lms on QA by switching to rolling
deploys. These change was at one point made manually but not committed
to the deployment repo. A recent change to the instance size then
unintentionally reset lms-qa to immutable deploys.

lms-qa and h-qa now use the same deployment settings, aside from
security group references, as shown by:

```
diff -u h/env-qa.yml lms/env-qa.yml
```

I also removed some commented-out configuration from the `h/env-qa.yaml` file.

### DIFF
--- a/h/env-qa.yml
+++ b/h/env-qa.yml
@@ -9,9 +9,7 @@ OptionSettings:
     StreamLogs: true
     RetentionInDays: "7"
   aws:elasticbeanstalk:command:
-    DeploymentPolicy: Rolling  #RollingWithAdditionalBatch #Immutable ## testing
-    # BatchSizeType: Fixed  ## testing
-    #BatchSize: 1 ## testing
+    DeploymentPolicy: Rolling
   aws:elasticbeanstalk:environment:
     EnvironmentType: LoadBalanced
     LoadBalancerType: application
@@ -19,7 +17,7 @@ OptionSettings:
   aws:elasticbeanstalk:environment:process:default:
     HealthCheckPath: /_status
   aws:elasticbeanstalk:healthreporting:system:
-    SystemType: enhanced #basic ## testing
+    SystemType: enhanced
   aws:ec2:vpc:
     Subnets: subnet-0f19e456,subnet-ee2c418b
     VPCId: vpc-bc4d91d9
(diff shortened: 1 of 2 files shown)


dmfine:
What we're doing in h-prod is the below - but not that I look at it I don't think that was captured in the config either :/   Sorry, I didn't understand how this worked back then. I'm making a note to make an issue to fix that. 

option_settings:
  aws:elasticbeanstalk:command:
    DeploymentPolicy: RollingWithAdditionalBatch
    BatchSizeType: Percentage
    BatchSize: 100

robertknight:
@dmfine - Thanks for checking re h prod's config. Can you take care of that? In the meantime, are you happy for me to merge this PR with the changes for lms-qa?

robertknight:
@dmfine - Ping. Can you take another look at this please when convenient?

dmfine:
What I meant to communicate so long ago was that the way we do it here should match the way we do it in h-prod. 

> option_settings:
> aws:elasticbeanstalk:command:
> DeploymentPolicy: RollingWithAdditionalBatch
> BatchSizeType: Percentage
> BatchSize: 100

I edited it directly on the branch, please double check my work and then it's got my seal of approval to merge

robertknight:
> I edited it directly on the branch, please double check my work and then it's got my seal of approval to merge

This confused me for a moment because this change, as explained in the description, was about making lms-qa work the same as h-qa. With the most recent commit `h/env-qa.yml` has now been changed (why?) and no longer matches `lms/env-qa.yml`.

My recollection is that the decision to use a different, faster deployment method for h-qa was an intentional choice some time ago. If you think we should reconsider then let's discuss - but I'd like to make that a separate change from what I was trying to do here which was speed up lms-qa deploys.

robertknight:
I discussed this with @dmfine on the devops sync yesterday. It seems like there was a misunderstanding here. I have reverted this branch such that it only makes a functional change to the lms QA environment. The change to `h/env-qa.yml` is purely about removing commented-out configuration.

robertknight:
Since this makes the same change that I've tested with Via recently, I'm going to go ahead and merge this and test with lms QA.

PR #46, opened 2019-09-19
Allow Via deploys to succeed even if some requests fail health checks

robertknight:
A large number of requests to Via proxying real-world web pages
naturally fail with 5xx responses due to issues such as:

 - Requests to invalid URLs triggered by incorrect URL rewriting in the
   front-end or non-rewritten URLs
 - Requests to inaccessible or blocked upstream URLs

Therefore in order to be able to deploy Via while the service is busy,
we need to tell AWS not to treat these as a sign of instance
unhealthiness.

This is a band-aid fix. We'll need to find something better as part of
finding a longer-term solution.

### DIFF
--- a/via/env-prod.yml
+++ b/via/env-prod.yml
@@ -15,6 +15,7 @@ OptionSettings:
     HealthCheckPath: /
   aws:elasticbeanstalk:healthreporting:system:
     SystemType: enhanced
+    HealthCheckSuccessThreshold: Degraded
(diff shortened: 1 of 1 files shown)


robertknight:
Slack discussion: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1568899575076900?thread_ts=1568899555.076600&cid=C4K6M7P5E

PR #53, opened 2019-10-30
Add IPTABLES rule to block docker container from accessing ec2metadata

dmfine:


dmfine:
This rule prevents access of ec2metadata IP address from within the docker container. This prevents exposing IAM credentials of any sort.

PR #54, opened 2019-10-30
simplify ebextension so that it simply inserts one iptables rule then…

dmfine:
… persists entire config to file


dmfine:
It's only in QA and the service is not production yet, so merging my own pr

PR #57, opened 2019-11-06
rotate the default ssh key

dmfine:
new key will be in passpack, passphrase held by dfine and a handful of others

### DIFF
--- a/h/env-prod.yml
+++ b/h/env-prod.yml
@@ -39,10 +39,10 @@ OptionSettings:
     SecurityGroups: sg-fdf9c19a,sg-a6073fc1
     IamInstanceProfile: aws-elasticbeanstalk-ec2-role
     InstanceType: m4.large
-    EC2KeyName: ops
+    EC2KeyName: ops-20191105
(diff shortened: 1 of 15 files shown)


h/env-prod.yml
```diff
@@ -39,10 +39,10 @@ OptionSettings:
     SecurityGroups: sg-fdf9c19a,sg-a6073fc1
     IamInstanceProfile: aws-elasticbeanstalk-ec2-role
     InstanceType: m4.large
-    EC2KeyName: ops
+    EC2KeyName: ops-20191105
   aws:autoscaling:asg:
     MaxSize: '12'
-    MinSize: '1'
+    MinSize: '4'
```
robertknight: Hi @dmfine - Can you revert this change as it isn't related to the SSH key change. If you still want to propose it for other reasons, can you make that part of a separate PR.

dmfine: Fixed

dmfine:
The key is in th 1password dev ops vault

PR #60, opened 2019-12-13
Use Health-based rolling updates for Via

robertknight:
Attempting to use "Immutable" rolling updates together with rolling
deployments appears not to be compatible. An env-sync task for
via/env-qa.yml failed with:

```
2019-12-13 16:53:14 UTC  ERROR  To enable immutable config and
application deployments together, you must set both the DeploymentPolicy
option (aws:elasticbeanstalk:command namespace) and the
RollingUpdateType option (aws:autoscaling:updatepolicy:rollingupdate
namespace) to Immutable.
```

See https://jenkins.hypothes.is/job/deployment/3104/console.

Since we want rolling deployments, change the rolling update mode to one
that is compatible with that.

### DIFF
--- a/via/env-qa.yml
+++ b/via/env-qa.yml
@@ -22,7 +22,7 @@ OptionSettings:
     ELBScheme: public
     AssociatePublicIpAddress: true
   aws:autoscaling:updatepolicy:rollingupdate:
-    RollingUpdateType: Immutable
+    RollingUpdateType: Health
     RollingUpdateEnabled: true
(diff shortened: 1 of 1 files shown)


robertknight:
Since the env-sync task is definitely not working in master, I'm going to merge this and see if it does work.

For reference, the relevant docs are at https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/using-features.rollingupdates.html.

PR #62, opened 2019-12-17
Fix h-qa rolling update type

robertknight:
The "Immutable" rolling update type is not compatible with the "Rolling"
deployment type. Change it to "Health" which matches lms-qa and via-qa.

The fact that these settings are incompatible presumably means that the
actual h-qa environment configuration was not in sync with h/env-qa.yml.

See also https://github.com/hypothesis/deployment/pull/60


robertknight:
CC @dmfine as this may come up with other environments in future.

robertknight:
I'm going to merge this because I know that what was in master previously was incorrect and the updated configuration here has been tested with lms-qa and via-qa.

PR #66, opened 2020-01-15
Add nginx logging into papertrail

hmstepanek:
Add nginx logs into papertrail.


hmstepanek:
Not sure if this is necessary in which case I'll close this see [slack thread](https://hypothes-is.slack.com/archives/C0LUWQQJJ/p1579047542006000).

PR #68, opened 2020-01-17
Change via deployment configuration

dmfine:
Change via instance type to t3.micro, use latest platform 2.14, and increase autoscale pool size  range to 10-20


dmfine:
I did get the changes through manually in AWS, since updating the platform had to happen before instance type change. I see there's a merge conflict, I'll address that and move this forward.

PR #73, opened 2020-02-04
Remove references to proxy-server

dmfine:
See https://github.com/hypothesis/via3/issues/17


seanh:
Clarification: proxy-server is **not** now called Via 3. Via 3 was previously called py-proxy. proxy-server is via**2**.hypothes.is

PR #74, opened 2020-02-05
Add more logs to PaperTrail

dmfine:
from host's /var/log and remove host's nginx since the nginx we care about is logging from within the container

### DIFF
--- a/via3/ebextensions/prod/papertrail.config
+++ b/via3/ebextensions/prod/papertrail.config
@@ -19,7 +19,8 @@ files:
     content: |
       files:
         - /var/log/eb-docker/containers/eb-current-app/*.log
-        - /var/log/nginx/*.log
+        - /var/log/*.log
+        - /var/log/secure
       hostname: ##DO_NOT_MODIFY-SETUP_WILL_REPLACE_THIS##
       destination:
         host: logs.papertrailapp.com


via3/ebextensions/prod/papertrail.config
```diff
@@ -19,7 +19,8 @@ files:
     content: |
       files:
         - /var/log/eb-docker/containers/eb-current-app/*.log
-        - /var/log/nginx/*.log
+        - /var/log/*.log
+        - /var/log/secure
```
seanh: Looking at the VM's `/var/log/` dir I see:

```
/var/log/
    /amazon/ssm/amazon-ssm-agent.log
    *.log (various files)
    cron (filename doesn't end with ".log")
    dmesg
    docker
    healthd/daemon.log
    lastlog
    maillog
    messages
    secure
    spooler
    wtmp
```

`/var/log/*.log` looks like it'll catch the top-level `*.log` files but won't catch:

1. Files that don't end with `.log`
2. Files in subdirs

Though I see you've added a separate rule for the `/var/log/secure` file.

Looking at this [example config](https://github.com/papertrail/remote_syslog2/blob/master/examples/log_files.yml.example.advanced) I think this might work:

```
files:
    - /var/log/*
    - /var/log/**/*
```

seanh: This also needs to be done for both prod and QA not just prod

seanh:
Closing in favor of: https://github.com/hypothesis/deployment/pull/75

via3/ebextensions/prod/papertrail.config
dmfine: I realize that we aren't capturing every single log on the host. It is however catching all the ones that are likely to contain anything of interest, and then a few more. Most of those are only writtn at the time of first boot. Some of the files that would match /var/log/**/* are binary format like lastlog. I think we should proceed with the original version.

PR #75, opened 2020-02-06
Send *everything* in /var/log/ to Papertrail

seanh:
Attempt to send all log files in /var/log/ to Papertrail, not just some of them. Trying this with Via 3 QA as a test.

### DIFF
--- a/via3/ebextensions/qa/papertrail.config
+++ b/via3/ebextensions/qa/papertrail.config
@@ -18,8 +18,7 @@ files:
     encoding: plain
     content: |
       files:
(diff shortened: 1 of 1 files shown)


seanh:
It looks like `**` should work btw, see https://github.com/papertrail/remote_syslog2/blob/master/examples/log_files.yml.example.advanced

dmfine:
As long as including the binary format files doesn't cause issues and we're not logging too much unnecessary stuff, that's fine by me. The original configuration I submitted was not a mistake, I just decided that /var/log/secure was the only log we might want which doesn't match *.log -- including more is probably okay though.

PR #83, opened 2020-02-11
Making via prod use rolling with batch deploy stategy

dmfine:
which is much faster than immutable. I could have sworn we already did this...

### DIFF
--- a/h/env-prod.yml
+++ b/h/env-prod.yml
@@ -25,7 +25,7 @@ OptionSettings:
     ELBScheme: public
     AssociatePublicIpAddress: true
   aws:autoscaling:updatepolicy:rollingupdate:
-    RollingUpdateType: Immutable
+    RollingUpdateType: Health
     RollingUpdateEnabled: true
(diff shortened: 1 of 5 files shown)


dmfine:
I have updated the EB environment by hand for now.

seanh:
:man_shrugging: I don't think we ever did get to the bottom of exactly what rolling-with-batch (and a batch size of 100%) actually does: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1579621988085400 So I can't review this.

I can just say that from a development perspective, it's unfortunately completely unacceptable for a failed deploy to ever take down production or leave production in an inconsistent or half-deployed state. We can't have developers deploying innocent PRs that then take down production Via (and by extension production LMS) in a way that they can't easily fix without ops knowledge.

So as long as that requirement is definitely met, then deployment changes are fine with me

dmfine:
Closing this because I'm addressing all the environment deploy strategies in a separate issue

PR #84, opened 2020-02-12
Remove public ips from instances

dmfine:
This PR will strip instances of public IP addresses so that they can only be accessed from the Load Balancer (or from a bastion host within the private network). This reduces our attack surface and prevents us from accidentally leaving a security group open to the public as has happened a few times recently. In rare cases where we need to connect to an instance directly, we can assign it an Elastic IP address and temporarily open the security group. 

There is a second commit to change the HealthCheck threshold for VIA to SEVERE, meaning it won't fail the deploy for DEGRADED and below. This is because VIA can get marked degraded in normal use if it's returning a lot of 400s. This config change has already been made manually in the console.

### DIFF
--- a/bouncer/env-prod.yml
+++ b/bouncer/env-prod.yml
@@ -23,7 +23,7 @@ OptionSettings:
     VPCId: vpc-bc4d91d9
     ELBSubnets: subnet-9bf985c3,subnet-3b21015f
     ELBScheme: public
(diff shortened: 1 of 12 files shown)


seanh:
:man_shrugging: I'm gonna opt out of reviewing this. I don't know of any reason _not_ to do it

PR #85, opened 2020-02-16
Change deploy types

dmfine:
deploy failed, seems like it needs to be rolling-with-batch for deploys and rolling for updates


dmfine:
I'm going to merge for the sake of time because it's only changing qa. The current config doesn't work.

PR #86, opened 2020-02-20
Update deployment strategy on prod environments.

dmfine:
The commit addresses two issues
1) updates the platform version for each environment to match what is live in production. This is necessary so that we don't regress to an older version when we sync environments. We're not updating everything to the latest platform yet, that will be a separate PR. 

2) it sets every prod environment to use the rolling-with-additional-batch code deployment strategy and immutable platform update strategy. Rolling-with-batch has been enabled for our critical services for many months already, and now we're expending to the lesser used services such as bouncer. It has the benefit of being much faster than immutable deploys. However we're keeping platform updates on the immutable strategy for production because we very rarely need to do that. Slow code deploys can result in downtime while we rollback bad updates, but platform updates aren't really subject to that.


dmfine:
closing to re-open as two PRs

PR #88, opened 2020-03-02
Increase LMS prod instance size to t3.small

seanh:
I'd like to increase LMS's instance size because (wild stab in the dark)
I think running out of memory might be a possible cause of DB connection
issues it seems to have been having. It currently uses 2-4 t2.nano's
which have only 512mb memory. That seems absurd -- we shouldn't be
trying to run our apps on 512mb. As a start I think we should increase
it to t3.small which has 2GB. I think that'll cost about $300 per year.

For comparison, h uses 4-12 m4.large's, legacy Via uses 10-20
t3.micro's,  Via 3 uses 4-12 t3.small's (and is currently just sitting
there, not even being used in production!).

### DIFF
--- a/lms/env-prod.yml
+++ b/lms/env-prod.yml
@@ -40,7 +40,7 @@ OptionSettings:
   aws:autoscaling:launchconfiguration:
     SecurityGroups: sg-fdf9c19a,sg-86756ee0
     IamInstanceProfile: aws-elasticbeanstalk-ec2-role
-    InstanceType: t2.nano
+    InstanceType: t3.small
     EC2KeyName: ops-20191105
   aws:autoscaling:asg:
     MaxSize: '4'


robertknight:
One issue that came up in the past when trying to switch to T3 instances was that the platform version, near the top of the config, needed to be updated. I don't know what the minimum version that supports T3 is, but h currently runs 2.14. This is currently on 2.12.

robertknight:
According to [EC2 Instance Pricing page](https://aws.amazon.com/ec2/pricing/on-demand/) a t3.small costs $0.0208/hour whereas a t2.nano costs $0.0058/hour. Over the course of a year (24*365 hours), with that's $131 per instance, so indeed $300 if we're running 2 instances most of the time.

robertknight:
I asked on Slack for the error details. From Sean:

> OperationalError: (psycopg2.OperationalError) SSL SYSCALL error: EOF detected

https://stackoverflow.com/questions/24130305/postgres-ssl-syscall-error-eof-detected-with-python-and-psycopg suggests that OOM may indeed by the issue.

seanh:
This has been synced to the LMS prod env

PR #94, opened 2020-05-07
Adding blank newrelic configuration for monitoring Via3

jon-betts:
This should enable host level monitoring of all (one) containers deployed for Via 3.

This is from: https://docs.newrelic.com/docs/infrastructure/new-relic-infrastructure/installation/install-infrastructure-agent-aws-elastic-beanstalk

----

I've just dumped the example files in for now as I'm not sure if it's a terrible idea to put in our license key in the clear.

### DIFF
--- a/lms/ebextensions/prod/newrelic.config
+++ b/lms/ebextensions/prod/newrelic.config
@@ -6,7 +6,7 @@ files:
     owner: root
     group: root
     content: |
-      license_key: YOUR_LICENSE_KEY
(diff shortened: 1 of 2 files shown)


jon-betts:
Rob found a good answer on Stack overflow about this: https://hypothes-is.slack.com/archives/CR3E3S7K8/p1588860280087200?thread_ts=1588858073.084700&cid=CR3E3S7K8

Not sure exactly what any of these files do, but I can pattern match with the best of 'em, so I've given this a shot.

PR #101, opened 2020-05-19
Add back Via3 NGINX monitoring qa 4

jon-betts:
An example of the output of `docker ps` from EB

  ```
 CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
   3e5e5bd34930        f670809f0e47        "/bin/sh -c '/usr/bi…"   17 hours ago        Up 17 hours         9083/tcp            quizzical_benz
```
Looks like nothing is preserved but the port.

Also trying to set some names from variables floating about. Maybe something will happen
if these values line up?


jon-betts:
While this is being developed I'm going to start merging these myself without review. It only effects QA and really nobody has any idea what will work. So it's throw lots of things at the wall and see what sticks time.

PR #104, opened 2020-07-06
Analytics dashboard config

indigobravo:
Hi Rob,

This is the deployment configuration for the analytics dashboard. It would be good to get some initial feedback on what has been done so far.

A very simple app has been created in the [analytics-dashboard repo](https://github.com/hypothesis/analytics-dashboard) and a Jenkins pipeline can be found [here](https://jenkins.hypothes.is/job/analytics-dashboard/)

I have no doubt there will be changes required to the app as Jon deploys his code, but this should get the deployment pipeline functional.

Let me know what you think?


analytics-dashboard/env-prod.yml
```diff
@@ -0,0 +1,55 @@
...
+  aws:elbv2:listener:443:
+    ListenerEnabled: true
+    SSLPolicy: ELBSecurityPolicy-TLS-1-2-2017-01
+    SSLCertificateArns: arn:aws:acm:us-west-1:964867326460:certificate/ef36b626-66e0-495e-813c-71a3b883f692
+    DefaultProcess: default
+    Protocol: HTTPS
+  aws:autoscaling:launchconfiguration:
+    SecurityGroups: sg-fdf9c19a
+    IamInstanceProfile: aws-elasticbeanstalk-ec2-role
+    InstanceType: t2.micro
```
robertknight, line 41: I'm not fully familiar with the differences between T2 and T3 instances, but I think we will probably be fine with "nano" instances for this app to start with, so maybe start with [t3.nano](https://aws.amazon.com/ec2/instance-types/t3/) and we can switch to a bigger instance later if we need it?

analytics-dashboard/env-prod.yml
```diff
@@ -0,0 +1,55 @@
...
+  aws:elasticbeanstalk:command:
+    DeploymentPolicy: RollingWithAdditionalBatch
+    BatchSizeType: Percentage
+    BatchSize: '100'
+  aws:elasticbeanstalk:environment:
+    EnvironmentType: LoadBalanced
+    LoadBalancerType: application
+    ServiceRole: aws-elasticbeanstalk-service-role
+  aws:elasticbeanstalk:environment:process:default:
+    HealthCheckPath: /
```
robertknight, line 17: Our apps typically have a `/_status` endpoint (eg. https://hypothes.is/_status) that is customised as appropriate for each app. I'd suggest we use that and ask Jon to add a suitable `_status` endpoint to the app.

analytics-dashboard/env-prod.yml
indigobravo: Thanks Rob. A new commit has been pushed and the suggest change has been made. I have also included ebextensions for 'bless' and 'papertrail'.

analytics-dashboard/env-prod.yml
indigobravo: Thanks Rob. A new commit has been pushed and the suggest change has been made. I have also included ebextensions for 'bless' and 'papertrail'.

PR #105, opened 2020-08-10
via3 iptables - block rfc1918 ip ranges

indigobravo:
Hi Guys,

This branch should address _most_ of what Jon is requesting in https://github.com/hypothesis/via3/issues/218

It will add the RFC1918 public IP address ranges to the DOCKER-USER iptables chain, and REJECT traffic from the container to those addresses.

**The update has only be done for QA. If this works I will create another branch for the prod update.**

What this doesn't address:

1. IPv6 - Our AWS VPC does not support IPv6 so I don't think that configuration is necessary.
2. Safe Ports. The [squid.conf](https://github.com/hypothesis/via/blob/master/conf/squid.conf) referenced in Jon's request defines a series of safe ports. Do we think these are necessary? I don't confess to understand exactly what via3 is doing, but my basic understand is it provides an HTTP proxy service for those that can not use the H client. If that is the case, shouldn't we only allow HTTP and HTTPS access?

```
acl SSL_ports port 443
acl Safe_ports port 80		# http
acl Safe_ports port 21		# ftp
acl Safe_ports port 443		# https
acl Safe_ports port 70		# gopher
acl Safe_ports port 210		# wais
acl Safe_ports port 1025-65535	# unregistered ports
acl Safe_ports port 280		# http-mgmt
acl Safe_ports port 488		# gss-http
acl Safe_ports port 591		# filemaker
acl Safe_ports port 777		# multiling http
```

**IPTABLES DOCKER-USER chain prior to this update**
```
[root@ip-10-1-3-250 ~]# iptables -L DOCKER-USER -n -v
Chain DOCKER-USER (1 references)
 pkts bytes target     prot opt in     out     source               destination
   12   720 REJECT     all  --  *      *       0.0.0.0/0            169.254.169.254      reject-with icmp-port-unreachable
1419K  522M RETURN     all  --  *      *       0.0.0.0/0            0.0.0.0/0
```

Hope this all makes sense. Let me know if you have any questions!

### DIFF
--- a/via3/ebextensions/qa/50_iptables_block_docker_metadata.config
+++ b/via3/ebextensions/qa/50_iptables_block_docker_metadata.config
@@ -5,5 +5,8 @@
 commands:
   10_insert_iptables_metadata_block:
     command: iptables --insert DOCKER-USER --destination 169.254.169.254 --jump REJECT
+    command: iptables --insert DOCKER-USER --destination 10.0.0.0/8 --jump REJECT
+    command: iptables --insert DOCKER-USER --destination 172.16.0.0/12 --jump REJECT
+    command: iptables --insert DOCKER-USER --destination 192.168.0.0/16 --jump REJECT
   20_persist_iptables:
     command: iptables-save > /etc/sysconfig/iptables
--- a/via3/ebextensions/qa/50_iptables_block_docker_metadata.config
+++ b/via3/ebextensions/qa/50_iptables_block_docker_metadata.config
@@ -4,9 +4,15 @@
 
 commands:
   10_insert_iptables_metadata_block:
+    # Block ec2metadata service
     command: iptables --insert DOCKER-USER --destination 169.254.169.254 --jump REJECT
+    # Block RFC 1918 private IP ranges
     command: iptables --insert DOCKER-USER --destination 10.0.0.0/8 --jump REJECT
     command: iptables --insert DOCKER-USER --destination 172.16.0.0/12 --jump REJECT
     command: iptables --insert DOCKER-USER --destination 192.168.0.0/16 --jump REJECT
+    # Block all accept TCP Ports 80 and 443
+    command: iptables --insert DOCKER-USER --protocol tcp --match tcp --match multiport ! --destination-ports 80,443 -j REJECT
+    # Allow UDP Port 53
+    command: iptables --insert DOCKER-USER --protocol udp --destination-port 53 --jump ACCEPT
   20_persist_iptables:
     command: iptables-save > /etc/sysconfig/iptables
--- a/via3/ebextensions/qa/50_iptables_block_docker_metadata.config
+++ b/via3/ebextensions/qa/50_iptables_block_docker_metadata.config
@@ -10,7 +10,7 @@ commands:
     command: iptables --insert DOCKER-USER --destination 10.0.0.0/8 --jump REJECT
     command: iptables --insert DOCKER-USER --destination 172.16.0.0/12 --jump REJECT
     command: iptables --insert DOCKER-USER --destination 192.168.0.0/16 --jump REJECT
-    # Block all accept TCP Ports 80 and 443
+    # Block all except TCP Ports 80 and 443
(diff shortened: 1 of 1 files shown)


jon-betts:
Via would need access to "anything that makes a webpage work" to do it's job. So if a site (like youtube.com) makes all sorts of calls to odd ports to get resources, we would want to allow that to some extent.

This also has to work with all sorts of potentially wonky corporate or internal sites, which I imagine are much more likely to be hosted on strange ports.

So I don't know, but I could guess, that we might have either have:

 * Taken a maximalist approach and only blocked certain known bad things
 * Added more and more to it over time to cover specific breaking examples

Given the targeted nature of some of the rules and the wide port band (1025-65535) it might be a bit of both.

In this case we're in quite a nice place to try this out and see what happens. We could restrict to just 80, 443 or something for Via 3 only, and see.

jon-betts:
It looks like this is a squid example file! So the chances are, the port choices above are probably not intentional. This might mean the only informational content there are the rules we've added ourselves.

robertknight:
> Via would need access to "anything that makes a webpage work" to do it's job. So if a site (like youtube.com) makes all sorts of calls to odd ports to get resources, we would want to allow that to some extent.

The vast majority of publicly accessible websites will do all their requests to ports 80 or 443 - trying to use other ports runs the risk of getting blocked by firewalls etc. in the way. I would like us to be conservative to begin with and only consider allowing other ports once we have concrete use cases.

via3/ebextensions/qa/50_iptables_block_docker_metadata.config
```diff
@@ -5,5 +5,8 @@
 commands:
   10_insert_iptables_metadata_block:
     command: iptables --insert DOCKER-USER --destination 169.254.169.254 --jump REJECT
+    command: iptables --insert DOCKER-USER --destination 10.0.0.0/8 --jump REJECT
```
robertknight, line 8: Could you add a comment above these 3 lines noting that they are RFC 1918 private IP ranges and a comment above the top one noting that it is the ec2metadata service - maybe just move the second sentence from the top of the file above that line.

robertknight:
> It will add the RFC1918 public IP address ranges

I presume you meant "private IP address ranges"?

indigobravo:
> > It will add the RFC1918 public IP address ranges
> 
> I presume you meant "private IP address ranges"?

Opps, my mistake!

indigobravo:
> > Via would need access to "anything that makes a webpage work" to do it's job. So if a site (like youtube.com) makes all sorts of calls to odd ports to get resources, we would want to allow that to some extent.
> 
> The vast majority of publicly accessible websites will do all their requests to ports 80 or 443 - trying to use other ports runs the risk of getting blocked by firewalls etc. in the way. There may indeed turn out to be good reasons to allow other ports, but I would like us to be conservative to begin with and only consider allowing other ports once we have concrete use cases. My rationale for this is that it is easier to open up new ports without causing regressions than to close them.

Additional rules have been include to block all ports other than:

* HTTP TCP 80
* HTTPS TCP 443
* DNS UDP 53

via3/ebextensions/qa/50_iptables_block_docker_metadata.config
indigobravo: Comments have been added and a new branch pushed.

via3/ebextensions/qa/50_iptables_block_docker_metadata.config
```diff
@@ -4,6 +4,15 @@
 
 commands:
   10_insert_iptables_metadata_block:
+    # Block ec2metadata service
     command: iptables --insert DOCKER-USER --destination 169.254.169.254 --jump REJECT
+    # Block RFC 1918 private IP ranges
+    command: iptables --insert DOCKER-USER --destination 10.0.0.0/8 --jump REJECT
+    command: iptables --insert DOCKER-USER --destination 172.16.0.0/12 --jump REJECT
+    command: iptables --insert DOCKER-USER --destination 192.168.0.0/16 --jump REJECT
+    # Block all accept TCP Ports 80 and 443
```
robertknight, line 13:
```suggestion
    # Block all except TCP Ports 80 and 443
```

PR #115, opened 2020-08-18
Via3 - iptables update prod + qa

indigobravo:
Here is a branch that will push the Via3 iptables firewall configuration that has been running in `qa` for the last few days over to `prod`. I have also included a small change to the comments in the `qa` version.

I would have preferred to find a better means to implement the solution, but I can not see a way to do this with Elastic Beanstalk. I have written up my findings here: https://github.com/hypothesis/playbook/issues/342#issuecomment-674009846

Let me know if you have any questions.


via3/ebextensions/prod/50_iptables.config
```diff
@@ -0,0 +1,46 @@
...
+files:
+  "/root/firewall.sh" :
+    mode: "000755"
+    owner: root
+    group: root
+    content: |
+      container_ip_one='172.17.0.2'
+      container_ip_two='172.17.0.3'
+      iptables --flush DOCKER-USER
+      iptables --insert DOCKER-USER --jump REJECT
```
robertknight, line 32: How do the priorities of these rules work? If for example TCP traffic matches both the `--protocol tcp --source $container_ip_one --jump ACCEPT` rule and the `--destination 10.0.0.0/8 --jump REJECT` rule, which wins?

via3/ebextensions/prod/50_iptables.config
```diff
@@ -0,0 +1,46 @@
...
+## 13. Accept ALL UDP DNS Traffic
+## 14. Save iptables rules
+##
+files:
+  "/root/firewall.sh" :
+    mode: "000755"
+    owner: root
+    group: root
+    content: |
+      container_ip_one='172.17.0.2'
```
robertknight, line 29: Is there an assumption here that the Docker container will be assigned one of these two IPs? Are there any changes we expect that might invalidate these assumptions in future?

via3/ebextensions/prod/50_iptables.config
```diff
@@ -0,0 +1,46 @@
+## Extension to deploy firewall script and execute
+##
+## FIREWALL SCRIPT EXPLAINED
```
robertknight, line 3: These comments re-state the `iptables` commands below, but I think it is more useful here to describe what the firewall config is trying to achieve, which is to prevent the Via 3 service from inadvertently serving any content from within our VPC when proxying a URL, such as private RFC 1918 IPs or the AWS metadata service. I would also suggest to note about where the `172.17.0.2` and `172.17.0.3.` values come from - I think these are just observed values that we've seen get used by the Docker daemon, so potentially they could change after a future OS or Docker version upgrade, which is something we'd need to be aware of?

via3/ebextensions/prod/50_iptables.config
indigobravo: The table I have produced here is slightly out of date because we are not dealing with ports anymore, but it should give you a good idea: https://github.com/hypothesis/playbook/issues/342#issue-678313730

Basically it is top to bottom. Whilst traversing the rules table a check will be compared against each rule if it matches the rule is applied and no other checks are completed. This could be an ACCEPT or REJECT.

We have a default rule at the bottom of the chain to REJECT anything that doesn't match any of the rules.

In essence, order matters a lot. Happy to go over this in more detail if you like. Maybe over a video call would be easier...

via3/ebextensions/prod/50_iptables.config
indigobravo: Hard one to answer... Based on my testing, albeit limited, I have not seen any other ip address being used. So I think we will be safe, but I can't be 100% sure. Technically, it could assign anything within the 172.17.0.0/12 subnet, but I can only imagine that would happen with large multi-container deployments.

We are kind of between a rock and a hard place here. I have searched for documentation defining how this works, but I have drawn a blank. I might be missing something, but I don't think so...

Another potential solutions, but I still see these as a hack... 

* ACCEPT traffic from a wider range, but where do we stop...
* Run a cron job that checks the docker containers IP once every minute to see if it has changed, and add a rule if it does.

A better solution if this fails might be to host Via3 in a DMZ completely on its own, away from anything that could be compromised.

via3/ebextensions/prod/50_iptables.config
indigobravo: Sensible suggestions.

via3/ebextensions/prod/50_iptables.config
robertknight: The approach taken here seems reasonable to me, as long as we understand the caveats. We can always change it if it stops working for any reason.

PR #116, opened 2020-08-27
NewRelic - Install infra agent

indigobravo:
This branch creates an `ebextension` to install NewRelic infra agent on analytics-dashboard-qa

### DIFF
--- a/analytics-dashboard/ebextensions/qa/100_new_relic_infra_agent.config
+++ b/analytics-dashboard/ebextensions/qa/100_new_relic_infra_agent.config
@@ -1,0 +1,37 @@
+# ebextension to install NewRelic infrastructure agent.
+#
+# License key configured via 'Environment property' in Elastic Beanstalk:
+# (Application > Configuration > Software > Environment properties)
+# 
+# Name of the environment property is: 'NEW_RELIC_LICENSE_KEY'
+#
+files:
+  "/root/new_relic_infra_agent_install.sh" :
+    mode: "000755"
+    owner: root
+    group: root
+    content: |
+      #!/bin/sh
+      echo "Obtaining NEW_RELIC_LICENSE_KEY"
+      key=`/opt/elasticbeanstalk/bin/get-config environment -k NEW_RELIC_LICENSE_KEY`
+      #
+      if [ -z "$key" ]
+      then
+        echo "NEW_RELIC_LICENSE_KEY is empty"
+        exit 1
+      else
+        echo "Creating license key file"
+        echo "license_key: $key" > /etc/newrelic-infra.yml
+        #
+        echo "Configuring infra agent yum repo"
+        curl -o /etc/yum.repos.d/newrelic-infra.repo https://download.newrelic.com/infrastructure_agent/linux/yum/el/6/x86_64/newrelic-infra.repo
+        #
+        echo "Updating yum cache for infra agent yum repo only"
+        yum -q makecache -y --disablerepo='*' --enablerepo='newrelic-infra'
+        #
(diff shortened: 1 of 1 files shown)


analytics-dashboard/ebextensions/qa/100_new_relic_infra_agent.config
```diff
@@ -0,0 +1,37 @@
...
+    owner: root
+    group: root
+    content: |
+      #!/bin/sh
+      echo "Obtaining NEW_RELIC_LICENSE_KEY"
+      key=`/opt/elasticbeanstalk/bin/get-config environment -k NEW_RELIC_LICENSE_KEY`
+      #
+      if [ -z "$key" ]
+      then
+        echo "key is empty"
```
robertknight, line 20: Can you change this to something slightly more descriptive in case the message gets emitted in a context where the source is not obvious? (eg. `NEW_RELIC_LICENSE_KEY environment variable is not set`).

analytics-dashboard/ebextensions/qa/100_new_relic_infra_agent.config
```diff
@@ -0,0 +1,37 @@
+# ebextension to install NewRelic infrastructure agent.
+#
+# License key configured via 'Environment property' in Elastic Beanstalk:
+# (Application > Configuration > Software > Environment properties)
+# 
+# Name of the environment property is: 'NEW_RELIC_LICENSE_KEY'
+#
+files:
+  "/root/infra_agent_install.sh" :
```
robertknight, line 9: I suggest to include New Relic in the filename somehow (`new_relic_infra_agent_install.sh`) to make it a little more obvious what "infra agent" is without having to look inside the file.

analytics-dashboard/ebextensions/qa/100_new_relic_infra_agent.config
```diff
@@ -0,0 +1,37 @@
...
+        exit 1
+      else
+        echo "Creating license key file"
+        echo "license_key: $key" > /etc/newrelic-infra.yml
+        #
+        echo "Configuring infra agent yum repo"
+        curl -o /etc/yum.repos.d/newrelic-infra.repo https://download.newrelic.com/infrastructure_agent/linux/yum/el/6/x86_64/newrelic-infra.repo
+        #
+        echo "Updating yum cache for infra agent yum repo only"
+        yum -q makecache -y --disablerepo='*' --enablerepo='newrelic-infra'
```
robertknight, line 30: Do any of these command line flags (eg. `--disablerepo`) have any effect beyond the duration of this command? ie. Could they cause issues with any EB extensions/commands which run later and assume that the default yum repos are available?

robertknight (approved):
LGTM with some minor questions and suggestions.

analytics-dashboard/ebextensions/qa/100_new_relic_infra_agent.config
indigobravo: No. The command excludes other repos from being updated when this command runs only. No effect on global configuration. It will only update cache for the newly added newrelic repo. It makes the process a little quicker to execute.

analytics-dashboard/ebextensions/qa/100_new_relic_infra_agent.config
indigobravo: That can be done

analytics-dashboard/ebextensions/qa/100_new_relic_infra_agent.config
indigobravo: That can be done

PR #130, opened 2020-09-17
Elastic Beanstalk - Via3 QA Ebextension - NginX proxy_buffer config.

indigobravo:
This branch delivers a new ebextension for via3 in qa only. The
extension configures a series of proxy parameters that enable us to
handle the large headers sent back by the likes of twitter and youtube.

Further details can be found here: https://github.com/hypothesis/via3/issues/250

### DIFF
--- a/via3/ebextensions/qa/60_proxy_buffers.config
+++ b/via3/ebextensions/qa/60_proxy_buffers.config
@@ -1,0 +1,16 @@
+# This ebextension modifies the default behaviour of NGINX as it is
+# shipped via Elastic Beanstalk. We are configuring a series of proxy
+# parameters that enable Via to handle large headers that the likes of
+# twitter and youtube send back.
+files:
+  "/etc/nginx/conf.d/01-proxy_buffers.conf" :
+    mode: "000644"
+    owner: root
+    group: root
+    content: |
+      proxy_buffer_size          128k;
+      proxy_buffers              4 256k;
+      proxy_busy_buffers_size    256k;
+container_commands:
+  nginx_reload:
+    command: "service nginx reload"


via3/ebextensions/qa/60_proxy_buffers.config
```diff
@@ -0,0 +1,16 @@
+# This ebextension modifies the default behaviour of NGINX as it is
+# shipped via Elastic Beanstalk. We are configuring a series of proxy
+# parameters that enable Via to handle large headers the likes of
```
robertknight, line 3:
```suggestion
# parameters that enable Via to handle large headers that the likes of
```

via3/ebextensions/qa/60_proxy_buffers.config
```diff
@@ -0,0 +1,16 @@
...
+# twitter and youtube send back.
+files:
+  "/etc/nginx/conf.d/01-proxy_buffers.conf" :
+    mode: "000644"
+    owner: root
+    group: root
+    content: |
+      proxy_buffer_size          128k;
+      proxy_buffers              4 256k;
+      proxy_busy_buffers_size    256k;
```
robertknight, line 13: Are there downsides to making the buffers much larger than the defaults? I would guess the main downside is that this will make nginx use more memory, which is probably not going to be an issue.

robertknight (approved):
Looks fine, with the caveat that I don't know much about nginx buffering.

via3/ebextensions/qa/60_proxy_buffers.config
indigobravo: You are correct in raising the point about memory consumption. However, I am not sure we can get more fine grained without testing the header size of all the sites we are likely to proxy, which could also change over time. Due to this, we might get into the situation where Via uses more memory than it previously use to. It also means we are vastly overspec for the majority of sites we will be working with.

The settings mean we will use 640K of RAM for each connection we are proxy. It will soon add up if we are serving many connections at the same time. Some rough sums....

1000 connections = 6.4GB
2000 connections = 12.8GB
3000 connections = 19.2GB
4000 connections = 25.6GB

via3/ebextensions/qa/60_proxy_buffers.config
indigobravo: Thanks, suggestion has been included in the code that will be committed.

PR #132, opened 2020-09-17
Add Papertrail logging for h-periodic

seanh:
It doesn't log much in normal usage, but I could imagine this being potentially useful for looking at error logging if debugging an h-periodic problem, or if we one day want more normal logging from h-periodic and change its log level.

### DIFF
--- a/h-periodic/ebextensions/prod/papertrail.config
+++ b/h-periodic/ebextensions/prod/papertrail.config
@@ -1,0 +1,195 @@
+# See http://help.papertrailapp.com/kb/hosting-services/aws-elastic-beanstalk/
+#
+# Usage:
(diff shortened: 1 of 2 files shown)


robertknight (approved):
For future reference, the two files added here are identical to the `papertrail.config` files in `h/ebextensions/prod/papertrail.config` and `h/ebextensions/qa/papertrail.config`. The only difference between the "qa" and "prod" configs is:

```diff
--- h-periodic/ebextensions/qa/papertrail.config        2020-10-07 15:20:00.000000000 +0100
+++ h-periodic/ebextensions/prod/papertrail.config      2020-10-07 15:20:00.000000000 +0100
@@ -23,7 +23,7 @@
       hostname: ##DO_NOT_MODIFY-SETUP_WILL_REPLACE_THIS##
       destination:
         host: logs.papertrailapp.com
-        port: 37197
+        port: 33163
         protocol: tls

   "/tmp/set-logger-hostname.sh":
```

PR #138, opened 2020-09-22
Switch metabase-prod to a larger instance type

robertknight:
Switching to an instance type with more memory should hopefully resolve a
deployment failure. Also switch from T2 => T3 instances for lower costs
for a given instance size. This required updating the platform version
to one that we know supports T3 instances.

### DIFF
--- a/metabase/env-qa.yml
+++ b/metabase/env-qa.yml
@@ -1,6 +1,6 @@
 AWSConfigurationTemplateVersion: 1.1.0.0
 Platform:
-  PlatformArn: arn:aws:elasticbeanstalk:us-west-1::platform/Docker running on 64bit Amazon Linux/2.7.0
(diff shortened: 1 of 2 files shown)


indigobravo (approved):
all looks good - approved.

metabase/env-qa.yml
```diff
@@ -38,7 +38,7 @@ OptionSettings:
   aws:autoscaling:launchconfiguration:
     SecurityGroups: sg-fdf9c19a,sg-816374e6
     IamInstanceProfile: aws-elasticbeanstalk-ec2-role
-    InstanceType: t2.micro
+    InstanceType: t3.micro
```
robertknight, line 41: I left the QA environment on a micro instance because the previous deployment worked fine with this instance size. I'd guess that is because there is a lot less happening on the QA one.

PR #139, opened 2020-09-22
Increase metabase qa instance size

robertknight:
Update metabase qa from a micro to a small instance as we have already
done for metabase prod.

This will hopefully resolve an issue during deployment:

```
Native memory allocation (mmap) failed to map 53780480 bytes for
committing reserved memory.
```

Discussion at https://discourse.metabase.com/t/minimum-requirements-to-host-metabase-tool/6712/3 suggests that 1GB is a sensible minimum so if there are multiple instances during a deployment, you really need 2GB.


robertknight:
Slack thread: https://hypothes-is.slack.com/archives/CR3E3S7K8/p1600763849038300

PR #140, opened 2020-10-08
Elastic Beanstalk Definition - ViaHTML3

indigobravo:
This update delivers a change to the Elastic Beanstalk definition for
ViaHTML3 in both QA and Production. The instance type has been changed
to t3.small.


robertknight (approved):
LGTM. For future reference, Slack thread discussing instance sizes: https://hypothes-is.slack.com/archives/CR3E3S7K8/p1602149044111400

PR #142, opened 2020-10-14
Ebextension - H-QA - 80_nginx_workers.config

indigobravo:
This branch delivers an ebextension that provisions a custom nginx.conf
for H-QA only.

We are introducing 2 changes:

1. Increase `worker_connections` to `1536`
2. Introducing the `worker_rlimit_nofile` parameter with a value of `1536`

These parameters are being tuned due to the recent performance problems we have been experiencing.


jon-betts:
It's 1024 for both.

jon-betts (approved):
I can't verify the ES portion of this, but the NGINX portion looks good. I think it's best to try deploying it to QA and then login and inspect the result.

Issue #145, opened 2020-10-20
Move ViaHTML directories to match the deployment plan


jon-betts:
The deployment repo needs some directories renamed for (https://github.com/hypothesis/via3/issues/283)

- viahtml3/ -> viahtml/ (so it matches the repo name)
- ebextensions/prod -> ebextensions/prod3 (so it matches the intended target)
- ebextensions/qa -> ebextensions/qa3 (so it matches the intended target)

jon-betts:
This is done, but hasn't taken effect yet, as we need the AWS updates.

PR #147, opened 2020-10-21
Remapping directories to match the deployment plan

jon-betts:
* Changing `viahtml3` to match the product `viahtml`
 * Creating 3 varients for the 3 environments (using symlink)

For: https://github.com/hypothesis/deployment/issues/145


viahtml/ebextensions/prod2
```diff
@@ -0,0 +1 @@
+prod
```
jon-betts, line 1: Doesn't look like here, but these are symlinks

indigobravo (approved):
approved.

jon-betts:
Again, we don't expect this to work until future work is done, but I'm committing it now so it doesn't get lost

PR #152, opened 2020-10-30
Make Via3 and ViaHTML papertrail logs the same as H and Via

indigobravo:
Here we have a branch that changes the papertrail logging for viahtml in both prod and qa. We are including the container and nginx logs.


jon-betts (approved):
This makes the config for Via HTML the same as H, which cuts down on the irrelevant info.

To give a flavour neither prod nor QA are doing anything but we've logged close to 150k messages from both

PR #166, opened 2020-11-20
EB Definition - Checkmate - Enabled auto-scaling.

indigobravo:
This branch updates the autoscaling configuration for Checkmate production. The configuration applied is the same as ViaHTML.


jon-betts (approved):
Looks good. Today I learned that having minimum 2 means that the two will run in different availability zones increasing reliability

PR #167, opened 2020-11-20
EBextension update - Removed access logs

indigobravo:
As per our discussion: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1605865607400500

We decided not to push access logs to papertrail for our proxy applications.

### DIFF
--- a/checkmate/ebextensions/prod/papertrail.config
+++ b/checkmate/ebextensions/prod/papertrail.config
@@ -19,7 +19,6 @@ files:
     content: |
       files:
         - /var/log/eb-docker/containers/eb-current-app/*.log
-        - /var/log/nginx/*.log
       hostname: ##DO_NOT_MODIFY-SETUP_WILL_REPLACE_THIS##
       destination:
         host: logs.papertrailapp.com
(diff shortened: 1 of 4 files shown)


via3/ebextensions/prod/papertrail.config
```diff
@@ -19,7 +19,6 @@ files:
     content: |
       files:
         - /var/log/eb-docker/containers/eb-current-app/*.log
-        - /var/log/nginx/*.log
```
jon-betts, line 22: Via 3 is a slightly different kettle of fish from Via and ViaHTML as it doesn't proxy HTML content (it just sends it to one of those to do the dirty work). So it should be fine to leave this in place.

jon-betts (changes requested):
This looks fine, except we can probably keep the Via 3 logs, as the traffic should be significantly lower.

via3/ebextensions/prod/papertrail.config
indigobravo: A new commit has been pushed. We are only removing the logs for:

* Checkmate (prod / qa)
* ViaHTML (prod / qa)

jon-betts (approved):
Looks good!

indigobravo:
Log filters have been implemented in papertail. Closing this PR.

PR #168, opened 2020-11-30
EBExtension - ViaHTML IP Tables

indigobravo:
This branch address the security issue raised by the SubGraph audit. https://github.com/hypothesis/private-issues/issues/56. It is the same code that we have deployed to Via3.

In essence we are using IP Tables to REJECT access from the docker container to RFC1918 IP Address space + The AWS metadata service.


robertknight:
Discussed in Slack here: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1606727638047900.

PR #180, opened 2021-02-09
EB Definition - ViaHTML

indigobravo:
This commit updates the environment definition for ViaHTML. We are moving to:

1. Min hosts online(MinSize): 4
2. Max hosts online(MaxSize): 10
3. CPU Scale out threshold(UpperThreshold): 45%
4. CPU Scale in threshold(LowerThreshold): 20%
5. A BreachDuration of: 4 mins (2x 2 min checks)

### DIFF
--- a/viahtml/env-prod.yml
+++ b/viahtml/env-prod.yml
@@ -43,15 +43,15 @@ OptionSettings:
     InstanceType: t3.small
     EC2KeyName: ops-20191105
   aws:autoscaling:asg:
-    MaxSize: '4'
-    MinSize: '2'
+    MaxSize: '10'
+    MinSize: '4'
   aws:autoscaling:trigger:
     MeasureName: CPUUtilization
     Statistic: Average
     Unit: Percent
     Period: '2'
     EvaluationPeriods: '2'
-    UpperThreshold: '60'
-    UpperBreachScaleIncrement: '1'
-    LowerThreshold: '30'
+    UpperThreshold: '45'
+    UpperBreachScaleIncrement: '2'
+    LowerThreshold: '20'
     LowerBreachScaleIncrement: '-1'


viahtml/env-prod.yml
```diff
@@ -43,15 +43,16 @@ OptionSettings:
...
   aws:autoscaling:trigger:
     MeasureName: CPUUtilization
     Statistic: Average
     Unit: Percent
     Period: '2'
     EvaluationPeriods: '2'
-    UpperThreshold: '60'
-    UpperBreachScaleIncrement: '1'
-    LowerThreshold: '30'
+    BreachDuration: '2'
```
robertknight, line 54: According to https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/command-options-general.html it looks like the default value, which we were using before, was 5 minutes. Is that correct?

viahtml/env-prod.yml
```diff
@@ -43,15 +43,16 @@ OptionSettings:
...
+    UpperThreshold: '45'
```
robertknight, line 55: 45 seems quite low to me to trigger a scale-up event. What was the rationale here?

viahtml/env-prod.yml
```diff
@@ -43,15 +43,16 @@ OptionSettings:
     InstanceType: t3.small
     EC2KeyName: ops-20191105
   aws:autoscaling:asg:
-    MaxSize: '4'
-    MinSize: '2'
+    MaxSize: '10'
+    MinSize: '4'
```
indigobravo, line 47: Run a minimum of 4 hosts at all times. This gives us some surge capacity.

viahtml/env-prod.yml
```diff
@@ -43,15 +43,16 @@ OptionSettings:
     InstanceType: t3.small
     EC2KeyName: ops-20191105
   aws:autoscaling:asg:
-    MaxSize: '4'
-    MinSize: '2'
+    MaxSize: '10'
```
indigobravo, line 46: Run a maximum of 10 hosts in this environment

viahtml/env-prod.yml
indigobravo, line 54: Change the breach duration to 2 minutes which is reduced from the default of 5 minutes. Enabling us to scale out quicker.

viahtml/env-prod.yml
indigobravo, line 55: Scale out when CPU breaches the 45% threshold. We have set this based on recent max CPU usage in NewRelic.

viahtml/env-prod.yml
```diff
@@ -43,15 +43,16 @@ OptionSettings:
...
     Statistic: Average
     Unit: Percent
     Period: '2'
     EvaluationPeriods: '2'
-    UpperThreshold: '60'
-    UpperBreachScaleIncrement: '1'
-    LowerThreshold: '30'
+    BreachDuration: '2'
+    UpperThreshold: '45'
+    UpperBreachScaleIncrement: '2'
```
indigobravo, line 56: Scale up by 2 hosts when the CPU metric is breached. This reduces multiple autoscaling events if we happen to see a large spike in usage.

viahtml/env-prod.yml
```diff
@@ -43,15 +43,16 @@ OptionSettings:
...
+    LowerThreshold: '20'
```
indigobravo, line 57: Reduce the lower threshold when we scale in and reduce the number of hosts we are running. This is now set to 20%

viahtml/env-prod.yml
robertknight, line 54: The [documentation for `aws:autoscaling:trigger`](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/command-options-general.html#command-options-general-autoscalingtrigger) has a note:

> Three options in this namespace determine how long a trigger's metric can remain beyond its defined limits before the trigger fires. These options are related as follows:
>
> BreachDuration = Period * EvaluationPeriods
> 
> The default values for these options (5, 5, and 1, respectively) satisfy this equation. If you specify inconsistent values, Elastic Beanstalk might modify one of the values so that the equation is still satisfied.

I read this as saying that you should either only specify two of these values and let the other be inferred, or specify all three in a way that is consistent with the above equation. Since we're already specifying `Period` and `EvaluationPeriods` it seems like we shouldn't need to specify `BreachDuration`.

robertknight (changes requested):
Reading the docs I think I found an issue with the way that `BreachDuration`, `Period` and `EvaluationPeriods` are set here.

viahtml/env-prod.yml
indigobravo: I have removed the `BreachDuration` parameter.

viahtml/env-prod.yml
robertknight: Using [this calculator](https://calculator.s3.amazonaws.com/index.html) I arrived at a cost increase of $36/month for the change from 2 => 4 instances by default. Is that right?

robertknight (approved):
OK, let's give this a try.

indigobravo:
You are correct about the pricing. This is the cost we are going to have to pay if we want to add surge capacity.

PR #181, opened 2021-02-26
EB - Update environment definition ViaHTML Prod

indigobravo:
This commit updates the EC2 instance type used to serve the ViaHTML application. Changing from `t3.small` to `t3.medium`.

Information on the reason for the change can be found here: https://github.com/hypothesis/playbook/issues/549

### DIFF
--- a/viahtml/env-prod.yml
+++ b/viahtml/env-prod.yml
@@ -40,7 +40,7 @@ OptionSettings:
   aws:autoscaling:launchconfiguration:
     SecurityGroups: sg-fdf9c19a
     IamInstanceProfile: aws-elasticbeanstalk-ec2-role
-    InstanceType: t3.small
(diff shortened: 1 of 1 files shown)


robertknight (approved):
Slack discussion: https://hypothes-is.slack.com/archives/CR3E3S7K8/p1614338020039100

It seems that the viahtml instances are frequently operating quite close to the 2GB limit at present. [See graph](https://hypothes-is.slack.com/archives/CR3E3S7K8/p1614338863040500?thread_ts=1614338020.039100&cid=CR3E3S7K8). Doubling the memory should give them more headroom. We haven't looked closely into where the memory is used. I think that would be a useful exercise to do, but we should give priority to retiring the legacy Via service first so that we only have one HTML proxying service whose operating characteristics we need to understand.

PR #182, opened 2021-03-02
EB - Environment definition updates for Via, Via3 and ViaHTML

indigobravo:
This commit brings our code in-line with the live configuration. It updates the environment definitions for the via proxies, and configures the environments to ignore 4xx responses received via the application and ELB. 

The method of defining these features seems a little vague in the documentation. I have made the update based on enhanced health details here: https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/health-enhanced-rules.html

I am not 100% sure if setting the `Enabled:` value to `false` enables or disables the check box as it is seen in the AWS console. So I have made my best guess. We may need need to make another commit if this has the opposite effect.

Screenshot 2021-03-02 at 14 16 00

### DIFF
--- a/via/env-prod.yml
+++ b/via/env-prod.yml
@@ -16,6 +16,16 @@ OptionSettings:
   aws:elasticbeanstalk:environment:process:default:
     HealthCheckPath: /
   aws:elasticbeanstalk:healthreporting:system:
+    ConfigDocument:
+      Rules:
+        Environment:
+          Application:
+            ApplicationRequests4xx:
+              Enabled: false
+          ELB:
+            ELBRequests4xx:
+              Enabled: false
+      Version: 1
     SystemType: enhanced
     HealthCheckSuccessThreshold: Severe
   aws:ec2:vpc:
--- a/via3/env-prod.yml
+++ b/via3/env-prod.yml
@@ -19,6 +19,16 @@ OptionSettings:
   aws:elasticbeanstalk:environment:process:default:
     HealthCheckPath: /_status
   aws:elasticbeanstalk:healthreporting:system:
+    ConfigDocument:
+      Rules:
+        Environment:
+          Application:
+            ApplicationRequests4xx:
+              Enabled: false
+          ELB:
+            ELBRequests4xx:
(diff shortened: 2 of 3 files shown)


via/env-prod.yml
```diff
@@ -16,6 +16,16 @@ OptionSettings:
   aws:elasticbeanstalk:environment:process:default:
     HealthCheckPath: /
   aws:elasticbeanstalk:healthreporting:system:
+    ConfigDocument:
```
indigobravo, line 19: Added two rules to disable the checking of application and load-balancer 4xx responses.

via3/env-prod.yml
```diff
@@ -19,6 +19,16 @@ OptionSettings:
   aws:elasticbeanstalk:environment:process:default:
     HealthCheckPath: /_status
   aws:elasticbeanstalk:healthreporting:system:
+    ConfigDocument:
```
indigobravo, line 22: Added two rules to disable the checking of application and load-balancer 4xx responses.

viahtml/env-prod.yml
```diff
@@ -19,6 +19,16 @@ OptionSettings:
   aws:elasticbeanstalk:environment:process:default:
     HealthCheckPath: /_status
   aws:elasticbeanstalk:healthreporting:system:
+    ConfigDocument:
```
indigobravo, line 22: Added two rules to disable the checking of application and load-balancer 4xx responses.

robertknight (approved):
The changes here look correct. I understand the confusion regarding the Elastic Beanstalk UI and its "Ignore " checkbox. To me the config here and documentation (see https://hyp.is/MsgYtIC7EeuTrlef9uztAw/docs.aws.amazon.com/elasticbeanstalk/latest/dg/health-enhanced-rules.html) make sense and it is the UI that seems either confusing or incorrect (more likely confusing).

I suggest we get this deployed and verify that it has the expected effect in practice.

Can you also add a note to the PR about why we're ignoring 4xx responses for Via. This seems obvious to us now but may not be so for a new dev in future.

PR #193, opened 2021-05-18
Rename via3 to via and delete legacyvia

seanh:


indigobravo:
The updates to the deployment config look good. 

However, I have a bit of a nagging doubt about how Elastic Beanstalk will update as I have never attempted this before. I have taken a look at the tooling. The update will not be performed by running the `sync-env` option in Jenkins. [eb-env-sync](https://github.com/hypothesis/deployment/blob/master/bin/eb-env-sync). It will be updated when running [eb-release](https://github.com/hypothesis/deployment/blob/master/bin/eb-release).

I think we are going to need to test this out and fix any problems we might run into along the way.

indigobravo (approved):
Let's give it a try

PR #194, opened 2021-06-04
Jenkinsfile updated to support deploying to multiple regions

indigobravo:
### Multi-Region Deployment Update

Here we have an update to the deployment workflow that enables multi-region deploys. This update can be merged and very little difference will be seen until we produce configs to deploy applications in Canada. However, it should be stated that this needs to be tested in a live environment which I will do once the PR has been landed.

**Compromise**

A compromise had to be made to avoid the process of rebuilding the entire suite of applications we have hosted in `us-west-1`, or having a completely different deployment workflow for Canada. The problem encountered related to filtering Elastic Beanstalk configurations that were designed for US or Canadian usage.

1. US Elastic Beanstalk configs will remain untouched and will follow our traditional naming scheme. An example would be:
    - h
    - lms
2. We need a method to filter Elastic Beanstalk configs that will be deployed to Canada. The solution proposed here changes our naming scheme to include a regional prefix . An example would be:
    - ca-h
    - ca-lms

**Updates**

The following updates have been made to the deployment Jenkinsfile.

1. The main stage has been renamed to `us-west-1`, and a when clause added to filter applications that DO NOT start `ca-`.
2. A new stage called `ca-central-1` has been introduce and a filter added to match applications that DO start with `ca-`.
3. The environment block used to define the AWS_DEFAULT_REGION environment variable has been copied to each of the stage blocks to enable the value to be varied based on the region being deployed.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -53,12 +53,16 @@ pipeline {
                description: 'Choose the application to deploy.')
     }
 
-    environment {
-        AWS_DEFAULT_REGION = 'us-west-1'
-    }
-
     stages {
-        stage('main') {
+        stage('us-west-1') {
+            when {
+                expression {
+                    !(params.APP ==~ /ca-.*/)
+                }
+            }
+            environment {
+                AWS_DEFAULT_REGION = 'us-west-1'
+            }
             steps {
                 script {
                     def label = "#${currentBuild.number} ${params.APP} " +
@@ -74,6 +78,33 @@ pipeline {
                 // parallel, as can deployments to different environments for
                 // the same app, but deployments to the same app and environment
                 // must execute serially.
+                lock(resource: "${params.APP}-${params.ENV}-deploy") {
+                    postSlack('start', params)
+                    withCredentials([[$class: 'AmazonWebServicesCredentialsBinding',
+                                      credentialsId: 'aws-elasticbeanstalk-jenkins']]) {
+                        sh 'bin/jenkins'
+                    }
+                }
+            }
+        }
+        stage('ca-central-1') {
+            when {
+                expression {
+                    params.APP ==~ /ca-.*/
+                }
+            }
+            environment {
+                AWS_DEFAULT_REGION = 'ca-central-1'
+            }
+            steps {
+                script {
+                    def label = "#${currentBuild.number} ${params.APP} " +
+                                "${params.ENV} ${params.TYPE}"
+                    currentBuild.displayName = label
+                }
+
+                // Deployment process in the ca-central-1 stage happen in the same
+                // manner as the us-west-1 stage.
                 lock(resource: "${params.APP}-${params.ENV}-deploy") {
                     postSlack('start', params)
                     withCredentials([[$class: 'AmazonWebServicesCredentialsBinding',


Jenkinsfile
```diff
@@ -53,12 +53,16 @@ pipeline {
                description: 'Choose the application to deploy.')
     }
 
-    environment {
```
indigobravo, line 56: The global `environment` block has been removed. We were only using it to define the `AWS_DEFAULT_REGION` environment variable which must now be varied based on the region being targeted by the deploy.

Jenkinsfile
```diff
@@ -53,12 +53,16 @@ pipeline {
...
-        AWS_DEFAULT_REGION = 'us-west-1'
-    }
-
     stages {
-        stage('main') {
```
indigobravo, line 61: The original `main` stage has been renamed to `us-west-1` to represent its new purpose, which is to deploy application configs hosted in `us-west-1`.

Jenkinsfile
```diff
@@ -53,12 +53,16 @@ pipeline {
...
+        stage('us-west-1') {
+            when {
```
indigobravo, line 58: A when clause has been introduced to exclude configs designed for Canada.

Jenkinsfile
```diff
@@ -53,12 +53,16 @@ pipeline {
...
+                expression {
+                    !(params.APP ==~ /ca-.*/)
+                }
+            }
+            environment {
```
indigobravo, line 63: Environment block inside the US stage that sets the `AWS_DEFAULT_REGION` environment variable to `us-west-1`

Jenkinsfile
```diff
@@ -83,6 +87,33 @@ pipeline {
                 }
             }
         }
+        stage('ca-central-1') {
```
indigobravo, line 90: New stage called `ca-central-1` has been introduced.

Jenkinsfile
```diff
@@ -83,6 +87,33 @@ pipeline {
...
+            when {
```
indigobravo, line 91: A when clause to include configs designed for Canada.

Jenkinsfile
```diff
@@ -83,6 +87,33 @@ pipeline {
...
+                expression {
+                    params.APP ==~ /ca-.*/
+                }
+            }
+            environment {
```
indigobravo, line 96: Environment block inside the CA stage that sets the `AWS_DEFAULT_REGION` environment variable to `ca-central-1`

robertknight:
I think I may have missed some previous discussion around this so apologies if I'm revisiting an old discussion. 

My expectation of how we would deploy apps to non-default regions is not as a different application/app but as a new environment for an existing application. In other words rather than having `params.APP` set to something like `ca-h` we would use the existing "h" app but have a new `ca-prod` environment. The deployment pipeline would then be something like:

```
Build => Test => QA deploy => [Wait for approval] => US Prod deploy
                                                  => Canada Prod deploy
```

indigobravo:
> My expectation of how we would deploy apps to non-default regions is not as a different application/app but as a new environment for an existing application. In other words rather than having `params.APP` set to something like `ca-h` we would use the existing "h" app but have a new `ca-prod` environment. The deployment pipeline would then be something like:
> 
> ```
> Build => Test => QA deploy => [Wait for approval] => US Prod deploy
>                                                   => Canada Prod deploy
> ```
> 
> Is this something that has already been considered, or do you see reasons we can't take this approach?

Hi Rob,

Thanks for the feedback, and apologies if this seems confusing.

The application workflows will operate similar to before. Here is an example of the Jenkins stage view for H.

Screenshot 2021-06-08 at 08 27 21

I am not proposing to change the `application` workflow as part of this PR, but when we get to that stage a `parallel` block will be added that will call the `deployment` workflow twice. Here is an example of a `parallel` block in the [Via Jenkinsfile](https://github.com/hypothesis/via/blob/master/Jenkinsfile)  

```
    stage("Deploy (prod)") {
	lock("prod deploy") {
	    parallel(
	        lms: {
		    sleep 2
		    deployApp(image: img, app: "lms-via", env: "prod")
		},
		public: {
		    deployApp(image: img, app: "via", env: "prod")
		}
	    )
	}
    }
```

This is used to call the `deployment` workflow twice to update both the `lms-via` and `via` Elastic Beanstalk environments, which is what I think your illustration depicts. The only difference here is both of the EB environments are in us-west-1.

---

The change here proposes two separate EB configs that run in different regions, but source the same Docker image. This provides us with the ability to customise resource definitions on a regional basis. 

In essence, we are introducing two stages blocks to the deployment workflow.

1. US-WEST-1
This block will only update Elastic Beanstalk environments based in us-west-1 that match a filter based on the `APP` name parameter. 

2. CA-CENTRAL-1
This block will only update Elastic Beanstalk environments based in ca-central-1 that match a filter based on the `APP` name parameter.

I am wondering if the term `APP` is becoming confusing here, because what we are really talking about is `Elastic Beanstalk environment`. A single GitHub and Docker repo will still be used to host the application code.



**Additional thoughts**

I have thought more about this since creating the PR. Another slightly cleaner solution could come from updating the structure of the deployment repo. However, this would also require updates to the scripts being called ([deployment/bin](https://github.com/hypothesis/deployment/tree/master/bin)). 

The layout of the deployment repo as it stands...
```
h\
  ebextensions\
  Dockerrun.aws.json
  env-prod.yml
  env-qa.yml
lms\
  ebextensions\
  Dockerrun.aws.json
  env-prod.yml
  env-qa.yml
```


Change proposed to store EB configs in a regional sub-directory.

```
EB Configs\
  US-WEST-1\ <- Default QA and Prod EB configs live here
    h\
      ebextensions\
      Dockerrun.aws.json
      env-prod.yml
      env-qa.yml
  CA-CENTRAL-1\ <- Prod EB configs live here
    h\
      ebextensions\
      Dockerrun.aws.json
      env-prod.yml
```

The benefit from the above change provides the same EB Environment name regardless of the region.

I hope this explains things a little better.

Let me know what you think?

indigobravo:
Closing in favour of: https://github.com/hypothesis/deployment/pull/195

PR #195, opened 2021-06-10
Updated Jenkinsfile to support a regional parameter.

indigobravo:
Here we have an update to the deployment Jenkinsfile to support deploying apps to multiple AWS regions. The workflow sets the default region to `us-west-1` so merging the change should not cause any issues with the application workflows already configured.

_Additional work after update has been merged_

- Add REGION parameter to the deployment pipeline GUI

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,18 +7,20 @@ def deployTypes = ['deploy', 'redeploy', 'sync-env'].join('\n')
 // The list of environments. It is assumed that each application has one of each
 // of these environments.
 def deployEnvironments = ['qa', 'prod'].join('\n')
+// The list of supported regions. Defaults to us-west-1.
+def deployRegions = ['us-west-1', 'ca-central-1'].join('\n')
 
 def postSlack(state, params) {
     def messages = [
-        'start': ['deploy': "Starting to deploy ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV}",
+        'start': ['deploy': "Starting to deploy ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV} (${params.REGION})",
                   'redeploy': "Starting re-deployment of ${params.APP} in ${params.ENV}",
-                  'sync-env': "Starting to synchronize the ${params.APP}-${params.ENV} environment"],
-        'success': ['deploy': "Successfully deployed ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV}",
-                    'redeploy': "Successfully re-deployed ${params.APP} in ${params.ENV}",
-                    'sync-env': "Successfully synchronized the ${params.APP}-${params.ENV} environment"],
-        'error': ['deploy': "Failed to deploy ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV}",
-                  'redeploy': "Failed to re-deploy ${params.APP} in ${params.ENV}",
-                  'sync-env': "Failed to synchronize the ${params.APP}-${params.ENV} environment"]
+                  'sync-env': "Starting to synchronize the ${params.APP}-${params.ENV} environment (${params.REGION})"],
+        'success': ['deploy': "Successfully deployed ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV} (${params.REGION})",
+                    'redeploy': "Successfully re-deployed ${params.APP} in ${params.ENV} (${params.REGION})",
+                    'sync-env': "Successfully synchronized the ${params.APP}-${params.ENV} environment (${params.REGION})"],
+        'error': ['deploy': "Failed to deploy ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV} (${params.REGION})",
+                  'redeploy': "Failed to re-deploy ${params.APP} in ${params.ENV} (${params.REGION})",
+                  'sync-env': "Failed to synchronize the ${params.APP}-${params.ENV} environment (${params.REGION})"]
     ]
     def colors = ['start': 'good', 'success': 'good', 'error': 'danger']
     slackSend color: colors[state], message: messages[state][params.TYPE]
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -7,18 +7,20 @@ def deployTypes = ['deploy', 'redeploy', 'sync-env'].join('\n')
 // The list of environments. It is assumed that each application has one of each
 // of these environments.
 def deployEnvironments = ['qa', 'prod'].join('\n')
+// The list of supported regions. Defaults to us-west-1.
+def deployRegions = ['us-west-1', 'ca-central-1'].join('\n')
```
indigobravo, line 11: Added a list that contains the AWS regions we support.

Jenkinsfile
```diff
@@ -7,18 +7,20 @@ def deployTypes = ['deploy', 'redeploy', 'sync-env'].join('\n')
...
-        'start': ['deploy': "Starting to deploy ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV}",
-                  'redeploy': "Starting re-deployment of ${params.APP} in ${params.ENV}",
-                  'sync-env': "Starting to synchronize the ${params.APP}-${params.ENV} environment"],
-        'success': ['deploy': "Successfully deployed ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV}",
-                    'redeploy': "Successfully re-deployed ${params.APP} in ${params.ENV}",
-                    'sync-env': "Successfully synchronized the ${params.APP}-${params.ENV} environment"],
-        'error': ['deploy': "Failed to deploy ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV}",
-                  'redeploy': "Failed to re-deploy ${params.APP} in ${params.ENV}",
-                  'sync-env': "Failed to synchronize the ${params.APP}-${params.ENV} environment"]
+        'start': ['deploy': "Starting ${params.REGION} deploy ${params.APP} ${params.APP_DOCKER_VERSION} to ${params.ENV}",
```
indigobravo, line 15: Updated the messages that get posted to Slack. They will now read...

```
Starting us-west-1 deploy h latest to prod
```
```
Starting us-west-1 re-deployment of h in prod
```
```
Starting us-west-1 synchronize of the h-prod environment
```
```
Successfully deployed us-west-1 h latest to prod
```
```
Successfully re-deployed us-west-1 h in prod
```
```
Successfully synchronized us-west-1 h-prod environment
```
```
Failed to deploy us-west-1 h latest to prod
```
```
Failed to re-deploy us-west-1 h in prod
```
```
Failed to synchronize us-west-1 h-prod environment
```

Jenkinsfile
```diff
@@ -48,13 +50,17 @@ pipeline {
                             'deploy. This is required if the selected ' +
                             'deployment type is `deploy`.',
                defaultValue: '')
+        choice(name: 'REGION',
```
indigobravo, line 53: Added a new choice parameter called `REGION` with a default value of `us-west-1`.

Jenkinsfile
```diff
@@ -48,13 +50,17 @@ pipeline {
...
+               description: 'Choose the region to deploy. Defaults to us-west-1',
+               defaultValue: 'us-west-1')
         choice(name: 'ENV',
                choices: deployEnvironments,
                description: 'Choose the application to deploy.')
     }
 
     environment {
-        AWS_DEFAULT_REGION = 'us-west-1'
+        AWS_DEFAULT_REGION = "${params.REGION}"
```
indigobravo, line 63: Updated the value for the `AWS_DEFAULT_REGION` environment variable. It is now set to the value of the `REGION` parameter.

Jenkinsfile
```diff
@@ -48,13 +50,16 @@ pipeline {
                             'deploy. This is required if the selected ' +
                             'deployment type is `deploy`.',
                defaultValue: '')
+        choice(name: 'REGION',
```
indigobravo, line 53: Add a new choice parameter called `REGION`, the value defaults to `us-west-1`.

Jenkinsfile
```diff
@@ -48,13 +50,16 @@ pipeline {
...
+               choices: deployRegions,
+               description: 'Choose the region to deploy. Defaults to us-west-1')
         choice(name: 'ENV',
                choices: deployEnvironments,
                description: 'Choose the application to deploy.')
     }
 
     environment {
-        AWS_DEFAULT_REGION = 'us-west-1'
+        AWS_DEFAULT_REGION = "${params.REGION}"
```
indigobravo, line 62: Updated the value for the `AWS_DEFAULT_REGION` environment variable. It is now set to the value of the `REGION` parameter.

Jenkinsfile
robertknight: Perhaps the messages might be a little easier to read if written like:

```
Starting to deploy h latest to prod (us-west-1)
```

Or:

```
Starting to deploy h latest to us-west-1/prod
```

In other words, keep it closer to the other part of the environment name (qa vs prod).

Jenkinsfile
```diff
@@ -48,13 +50,16 @@ pipeline {
                             'deploy. This is required if the selected ' +
                             'deployment type is `deploy`.',
                defaultValue: '')
+        choice(name: 'REGION',
+               choices: deployRegions,
+               description: 'Choose the region to deploy. Defaults to us-west-1')
```
robertknight, line 55: I was expecting to see `defaultValue: 'us-west-1'` here. What happens if you don't specify a `defaultValue` for an optional argument, does that result in an error or do you get some kind of null value?

robertknight (commented):
I had a query related to the default value for the `REGION` argument and a suggestion regarding the message formatting.

Jenkinsfile
indigobravo: Thanks for the feedback - Messages have been updated. I went with your first suggestion
```
Starting to deploy h latest to prod (us-west-1)
```

Jenkinsfile
indigobravo: It threw me a little when I first looked at it. If I am understanding the syntax correctly, there is no `defaultValue` for a `choice` parameter. It uses the first option in the list of values being provided to it. 

In our case the  `deployRegions` list is being provided to the `choices` option. The list contains:
```
['us-west-1', 'ca-central-1']
```
So the default is `us-west-1`

I have tested this behaviour in a local Jenkins instance I have running at it works as I have described.

The `string` parameter we are using to capture `APP_DOCKER_VERSION` does have a `defaultValue` which in our scenario is set to `''`, which I believe is providing a blank / null value.

More info on Jenkins parameters can be found here: https://www.jenkins.io/doc/book/pipeline/syntax/#parameters

robertknight: Ah, OK. Thanks for the clarification.

robertknight (approved):
LGTM

PR #197, opened 2021-06-10
Deployment tooling update - bin/jenkins

indigobravo:
The bin/jenkins script has been updated to handle multi-region deployments.

---

### DIFF
--- a/bin/jenkins
+++ b/bin/jenkins
@@ -14,10 +14,14 @@
 # name of the environment in Elastic Beanstalk will be prefixed with the name of
 # the application.
 : ${ENV:=}
+# The target AWS region
+: ${REGION:=}
 
 PATH="$(dirname "$0"):${PATH}"
 
 set -eu
+
+export AWS_DEFAULT_REGION="$REGION"
 
 abort () {
     echo "Error:" "$@" >&2
@@ -33,26 +37,30 @@ if [ -z "$ENV" ]; then
     abort "cannot proceed unless \$ENV is specified"
 fi
 
+if [ -z "$REGION" ]; then
+    abort "cannot proceed unless \$REGION is specified"
+fi
+
 if [ "$TYPE" = "deploy" ]; then
   if [ -z "$APP_DOCKER_VERSION" ]; then
     abort "cannot proceed with deploy unless \$APP_DOCKER_VERSION is specified"
   fi
 
-  version_label=$(eb-release "$APP" "$APP_DOCKER_VERSION" "$ENV")
-  eb-deploy "$APP" "$ENV" "$version_label"
+  version_label=$(eb-release "$APP" "$APP_DOCKER_VERSION" "$ENV" "$REGION")
+  eb-deploy "$APP" "$ENV" "$version_label" "$REGION"
 
 elif [ "$TYPE" = "redeploy" ]; then
   if [ -n "$APP_DOCKER_VERSION" ]; then
     abort "do not specify \$APP_DOCKER_VERSION for a redeploy"
   fi
 
-  version_label=$(eb-env-version "$APP" "$ENV")
-  eb-deploy "$APP" "$ENV" "$version_label"
+  version_label=$(eb-env-version "$APP" "$ENV" "$REGION")
+  eb-deploy "$APP" "$ENV" "$version_label" "$REGION"
 
 elif [ "$TYPE" = "sync-env" ]; then
-  eb-env-sync "$APP" "$ENV"
+  eb-env-sync "$APP" "$ENV" "$REGION"
 else
   abort "Don't know how to handle deployment type \"$TYPE\""
 fi
 
-eb-env-wait "$APP" "$ENV"
+eb-env-wait "$APP" "$ENV" "$REGION"


bin/jenkins
```diff
@@ -14,6 +14,8 @@
 # name of the environment in Elastic Beanstalk will be prefixed with the name of
 # the application.
 : ${ENV:=}
+# The target AWS region
```
indigobravo, line 17: I have copied the format of the original script here...

If I am understanding this correct, the `:` basically does nothing, and sets `REGION` to a null value if it does not exist.

bin/jenkins
```diff
@@ -33,24 +35,28 @@ if [ -z "$ENV" ]; then
     abort "cannot proceed unless \$ENV is specified"
 fi
 
+if [ -z "$REGION" ]; then
```
indigobravo, line 38: Check if the `REGION` variable is null and exit.

bin/jenkins
```diff
@@ -33,24 +35,28 @@ if [ -z "$ENV" ]; then
...
+fi
+
 if [ "$TYPE" = "deploy" ]; then
   if [ -z "$APP_DOCKER_VERSION" ]; then
     abort "cannot proceed with deploy unless \$APP_DOCKER_VERSION is specified"
   fi
 
   version_label=$(eb-release "$APP" "$APP_DOCKER_VERSION" "$ENV")
-  eb-deploy "$APP" "$ENV" "$version_label"
+  eb-deploy "$APP" "$ENV" "$version_label" "$REGION"
```
indigobravo, line 48: Pass `REGION` to `eb-deploy`

bin/jenkins
```diff
@@ -33,24 +35,28 @@ if [ -z "$ENV" ]; then
...
 
 elif [ "$TYPE" = "redeploy" ]; then
   if [ -n "$APP_DOCKER_VERSION" ]; then
     abort "do not specify \$APP_DOCKER_VERSION for a redeploy"
   fi
 
   version_label=$(eb-env-version "$APP" "$ENV")
-  eb-deploy "$APP" "$ENV" "$version_label"
+  eb-deploy "$APP" "$ENV" "$version_label" "$REGION"
```
indigobravo, line 56: Pass `$REGION` to `eb-deploy`

bin/jenkins
```diff
@@ -33,24 +35,28 @@ if [ -z "$ENV" ]; then
...
 
 elif [ "$TYPE" = "sync-env" ]; then
-  eb-env-sync "$APP" "$ENV"
+  eb-env-sync "$APP" "$ENV" "$REGION"
```
indigobravo, line 59: Pass `$REGION` to `eb-env-sync`

bin/jenkins
robertknight: That matches my understanding. `:` is a no-op command that allows `${VAR:=val}` to be executed. I guess this coding pattern is to allow known variables to be "registered" so they don't trigger unknown variable errors if not set.

robertknight (approved):
LGTM

Edit: I realized there is a hazard here where the aws commands could end up operating against a different region than what is specified via `REGION`. We need to prevent that somehow.

bin/jenkins
```diff
@@ -14,11 +14,15 @@
...
 # the application.
 : ${ENV:=}
+# The target AWS region
+: ${REGION:=}
 
 PATH="$(dirname "$0"):${PATH}"
 
 set -eu
 
+export AWS_DEFAULT_REGION="$REGION"
```
indigobravo, line 24: Purposely setting the `AWS_DEFAULT_REGION` environment variable before invoking any commands.

bin/jenkins
```diff
@@ -33,26 +37,30 @@ if [ -z "$ENV" ]; then
...
+fi
+
 if [ "$TYPE" = "deploy" ]; then
   if [ -z "$APP_DOCKER_VERSION" ]; then
     abort "cannot proceed with deploy unless \$APP_DOCKER_VERSION is specified"
   fi
 
-  version_label=$(eb-release "$APP" "$APP_DOCKER_VERSION" "$ENV")
-  eb-deploy "$APP" "$ENV" "$version_label"
+  version_label=$(eb-release "$APP" "$APP_DOCKER_VERSION" "$ENV" "$REGION")
```
indigobravo, line 49: Pass `$REGION` to `eb-release`

bin/jenkins
```diff
@@ -33,26 +37,30 @@ if [ -z "$ENV" ]; then
...
 
 elif [ "$TYPE" = "sync-env" ]; then
-  eb-env-sync "$APP" "$ENV"
+  eb-env-sync "$APP" "$ENV" "$REGION"
 else
   abort "Don't know how to handle deployment type \"$TYPE\""
 fi
 
-eb-env-wait "$APP" "$ENV"
+eb-env-wait "$APP" "$ENV "$REGION""
```
indigobravo, line 66: Pass `$REGION` to `eb-env-wait`

bin/jenkins
indigobravo: Thanks for the clarification.

bin/jenkins
```diff
@@ -33,24 +37,28 @@ if [ -z "$ENV" ]; then
...
+  eb-deploy "$APP" "$ENV" "$version_label" "$REGION"
 
 elif [ "$TYPE" = "redeploy" ]; then
   if [ -n "$APP_DOCKER_VERSION" ]; then
     abort "do not specify \$APP_DOCKER_VERSION for a redeploy"
   fi
 
-  version_label=$(eb-env-version "$APP" "$ENV")
-  eb-deploy "$APP" "$ENV" "$version_label"
+  version_label=$(eb-env-version "$APP" "$ENV" "$REGION")
```
indigobravo, line 57: Passing `$REGION` to the `eb-env-version` script

bin/jenkins
```diff
@@ -33,26 +37,30 @@ if [ -z "$ENV" ]; then
...
 
 elif [ "$TYPE" = "sync-env" ]; then
-  eb-env-sync "$APP" "$ENV"
+  eb-env-sync "$APP" "$ENV" "$REGION"
 else
   abort "Don't know how to handle deployment type \"$TYPE\""
 fi
 
-eb-env-wait "$APP" "$ENV"
+eb-env-wait "$APP" "$ENV" "$REGION"
```
indigobravo, line 66: Pass `$REGION` to `eb-env-wait`

PR #201, opened 2021-06-11
Deployment tooling update - bin/eb-env-exists

indigobravo:
The bin/eb-env-exists script has been updated to handle multi-region
deployments.

### DIFF
--- a/bin/eb-env-exists
+++ b/bin/eb-env-exists
@@ -1,13 +1,13 @@
 #!/bin/sh
 
 usage () {
-    echo "Usage: $(basename "$0") <APP> <ENV>"
+    echo "Usage: $(basename "$0") <APP> <ENV> <REGION>"
     echo
     echo "Check if an Elastic Beanstalk environment already exists. Exits zero"
     echo "when the environment exists, and non-zero otherwise."
 }
 
-if [ "$#" -ne 2 ]; then
+if [ "$#" -ne 3 ]; then
     usage >&2
     exit 1
 fi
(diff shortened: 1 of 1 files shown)


bin/eb-env-exists
```diff
@@ -1,13 +1,13 @@
...
 usage () {
-    echo "Usage: $(basename "$0") <APP> <ENV>"
+    echo "Usage: $(basename "$0") <APP> <ENV> <REGION>"
     echo
     echo "Check if an Elastic Beanstalk environment already exists. Exits zero"
     echo "when the environment exists, and non-zero otherwise."
 }
 
-if [ "$#" -ne 2 ]; then
+if [ "$#" -ne 3 ]; then
```
indigobravo, line 10: Increased mandatory parameters to 4 now we are passing in `$REGION`

bin/eb-env-exists
```diff
@@ -16,10 +16,12 @@ set -eu
 
 APP=$1
 ENV=$2
+REGION=$3
```
indigobravo, line 19: Set `$REGION` to $3

bin/eb-env-exists
```diff
@@ -16,10 +16,12 @@ set -eu
...
 
 count=$(aws elasticbeanstalk describe-environments \
         --application-name "$APP" \
         --environment-names "${APP}-${ENV}" \
         --query 'length(Environments)' \
-        --output text)
+        --output text) \
+        --region "$REGION"
```
indigobravo, line 26: Added `--region` parameter to aws command

robertknight: The `--region` option has been added outside of the sub-shell `$(...)` here so it won't be passed to the `aws` command.

robertknight (changes requested):
The change looks good, except for an error in how the `--region` argument is passed. Good to merge once that is resolved.

bin/eb-env-exists
indigobravo: Thanks for spotting this, a new update has been pushed.

robertknight (approved):

PR #205, opened 2021-06-11
Deployment tooling update - bin/eb-env-version

indigobravo:
The bin/eb-env-version script has been updated to support multi-region
deployments

### DIFF
--- a/bin/eb-env-version
+++ b/bin/eb-env-version
@@ -14,12 +14,13 @@ import boto3
 parser = argparse.ArgumentParser()
 parser.add_argument('app')
 parser.add_argument('env')
+parser.add_argument('region')
 
 
 def main():
     args = parser.parse_args()
 
-    client = boto3.client('elasticbeanstalk')
+    client = boto3.client('elasticbeanstalk', region_name=args.region)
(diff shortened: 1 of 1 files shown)


bin/eb-env-version
```diff
@@ -14,12 +14,13 @@ import boto3
 parser = argparse.ArgumentParser()
 parser.add_argument('app')
 parser.add_argument('env')
+parser.add_argument('region')
```
indigobravo, line 17: Added region parameter

bin/eb-env-version
```diff
@@ -14,12 +14,13 @@ import boto3
...
 
 
 def main():
     args = parser.parse_args()
 
-    client = boto3.client('elasticbeanstalk')
+    client = boto3.client('elasticbeanstalk', region_name=args.region)
```
indigobravo, line 23: Specifying `region_name` in the `boto3.client` definition which is set to the value of the `region` parameter.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html

robertknight (approved):

PR #206, opened 2021-06-11
Deployment tooling update - bin/eb-task-wait

indigobravo:
The bin/eb-task-wait script has been updated to support multi-region
deployments

### DIFF
--- a/bin/eb-task-wait
+++ b/bin/eb-task-wait
@@ -22,6 +22,7 @@ IN_PROGRESS_STATUSES = ['Pending', 'In Progress', 'Delayed']
 parser = argparse.ArgumentParser()
 parser.add_argument('command_id')
 parser.add_argument('instance_id')
+parser.add_argument('region')
 
 
 def main():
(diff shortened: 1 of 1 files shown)


bin/eb-task-wait
```diff
@@ -22,6 +22,7 @@ IN_PROGRESS_STATUSES = ['Pending', 'In Progress', 'Delayed']
 parser = argparse.ArgumentParser()
 parser.add_argument('command_id')
 parser.add_argument('instance_id')
+parser.add_argument('region')
```
indigobravo, line 25: Added `region` parameter.

bin/eb-task-wait
```diff
@@ -30,7 +31,7 @@ def main():
 
 
 def loop(args):
-    client = boto3.client('ssm')
+    client = boto3.client('ssm', region_name=args.region)
```
indigobravo, line 34: Specifying `region_name` in the `boto3.client` definition which is set to the value of the `region` parameter.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html

robertknight (approved):

PR #214, opened 2021-06-24
Created iamsync ebextension

indigobravo:
This commit removes the bless extension and delivers the initial
extension we will use to install iamsync on EC2 hosts provisioned
by Elastic Beanstalk.


analytics-dashboard/ebextensions/qa/90_iamsync.config
```diff
@@ -0,0 +1,59 @@
...
+  01python-venv
+    command: "python3 -m venv /root/venv"
+  02create-dir:
+    command: "mkdir /root/bin"
+  03move-iamsync:
+    command: "mv /tmp/iamsync.py /root/bin/iamsync.py"
+  04pip-install:
+    command: "source /root/bin/venv/bin/activate ; pip install wheel boto3 pyyaml"
+  05create-crontab:
+    command: "crontab -u root /tmp/cronjobs.txt"
```
robertknight, line 59: Pardon my ignorance here: does this append to or replace the existing cron jobs? ie. what happens if another ebextension is also trying to create cron commands?

analytics-dashboard/ebextensions/qa/90_iamsync.config
```diff
@@ -0,0 +1,59 @@
...
+  "/root/iam.yml":
+    mode: 
+    mode: "000600"
+    owner: root
+    group: root
+    content: |
+      iamsync:
+        - iam_group: engineering
+          sudo_rule: "ALL=(ALL) NOPASSWD:ALL"
+          local_gid: 1025
```
robertknight, line 46: Where does this group ID come from?

analytics-dashboard/ebextensions/qa/90_iamsync.config
```diff
@@ -0,0 +1,59 @@
...
...Zy5pbmZvKGYib3JwaGFuIGFjY291bnQgaWRlbnRpZmllZCAoe3VzZXJuYW1lfSkiKQogICAgICAgICAgICAgICAgbGludXhfdXNlcl9kZWxldGUodXNlcm5hbWUpCiAgICBleGNlcHQgRXhjZXB0aW9uIGFzIGU6CiAgICAgICAgcmFpc2UgRXhjZXB0aW9uKCJvcnBoYW5fYWNjb3VudF9jbGVhbnVwIGZhaWxlZCIpIGZyb20gZQoKCmlmIF9fbmFtZV9fID09ICJfX21haW5fXyI6CiAgICB0cnk6CiAgICAgICAgbG9nZ2luZy5pbmZvKCJTdGFydGluZyBpYW1zeW5jIikKICAgICAgICBpYW1fdGVzdCgpCiAgICAgICAgY29uZmlnX2xpc3QgPSByZWFkX2NvbmZpZygpCiAgICAgICAgZm9yIGNvbmZpZyBpbiBjb25maWdfbGlzdDoKICAgICAgICAgICAgZ3JvdXAgPSBjb25maWdbImlhbV9ncm91cCJdCiAgICAgICAgICAgIGdpZCA9IGNvbmZpZ1sibG9jYWxfZ2lkIl0KICAgICAgICAgICAgc3Vkb19ydWxlID0gY29uZmlnWyJzdWRvX3J1bGUiXQogICAgICAgICAgICBpYW1fdXNlcl9pbmZvID0gaWFtX3VzZXJfcXVlcnkoZ3JvdXApCiAgICAgICAgICAgIGlmIGlhbV91c2VyX2luZm8gaXMgbm90IE5vbmU6CiAgICAgICAgICAgICAgICBsaW51eF9ncm91cF92YWxpZGF0ZShncm91cCwgZ2lkKQogICAgICAgICAgICAgICAgc3Vkb19ydWxlX3ZhbGlkYXRlKHN1ZG9fcnVsZSwgZ3JvdXApCiAgICAgICAgICAgICAgICBsaW51eF91c2VyX3ZhbGlkYXRlKGlhbV91c2VyX2luZm8sIGdyb3VwKQogICAgICAgIG9ycGhhbl9hY2NvdW50X2NsZWFudXAoKQogICAgICAgIGxvZ2dpbmcuaW5mbygiQ29tcGxldGVkIGlhbXN5bmMiKQogICAgZXhjZXB0IEV4Y2VwdGlvbiBhcyBlOgogICAgICAgIGxvZ2dpbmcuZXJyb3IoZiJfX21haW5fXyAge2V9IikKICAgICAgICByYWlzZSBTeXN0ZW1FeGl0KDEpCg==
```
robertknight, line 33: Hmm - it's a bit annoying to have this in base64 form in the repo. Would it be possible to either have the file in this repo but in plain text (diffable) form or alternatively download it from somewhere (eg. an S3 bucket). I feel like I might have asked a similar question before on a different PR. Apologies if this is rehashing a previous discussion.

analytics-dashboard/ebextensions/qa/90_iamsync.config
```diff
@@ -0,0 +1,59 @@
...
+files:
+  # Create a template crontab that gets sourced in the commands section
+  "/tmp/cronjobs.txt":
+    mode: "000755"
+    owner: root
+    group: root
+    content: |
+      SHELL=/bin/bash
+      PATH="venv/bin:/usr/sbin:/bin"
+      5,35 * * * * sleep ${RANDOM: -1} ; python bin/iamsync.py
```
robertknight, line 22: To check my understanding: `sleep ${RANDOM: -1}` sleeps for a random duration between 0 and 9 seconds? What is the purpose of this?

robertknight (commented):
I had a few questions on some details, mostly to help with my understanding of a few details. A more general question I have is whether we'll get any kind of notification or alert if the cron job fails to run for any reason?

analytics-dashboard/ebextensions/qa/90_iamsync.config
indigobravo: Good question - I am still testing this out and deciding on the best method to install. Your question is very valid, and has made me think... In its current form the method proposed here will wipe out any existing entries.

analytics-dashboard/ebextensions/qa/90_iamsync.config
indigobravo: The ID is randomly generated. All I am really looking to do here is leave a decent amount of space from standard system generated IDs

analytics-dashboard/ebextensions/qa/90_iamsync.config
indigobravo: I with you here. It would be nice to have it in one location where it can be install easily in both contexts that we use:

1. Installed via an ansible role in the playbook repo
2. Installed via an ebextension in the deployment repo

I am tempted to try the solution I have used in Canada where a deployment key is used to `pull` a copy of the playbook repo where the iamsync script lives.

analytics-dashboard/ebextensions/qa/90_iamsync.config
indigobravo: Your understanding is correct. It was added to avoid a situation where AWS could potentially rate limit us because too many API requests are being made at the same time.

I have also chosen to run the task at `05` and `35` minutes past the hour. This differs to the ansible implementation where it runs at `00` and `30` minutes past.

indigobravo:
> I had a few questions on some details, mostly to help with my understanding of a few details. A more general question I have is whether we'll get any kind of notification or alert if the cron job fails to run for any reason?

This is still the initial phase of testing this out and getting it working. However, you raise a valid point. As it stands an alert will not be generated if cron fails. However, once this is working, I am intending on pushing the iamsync.log file to papertrail where we can setup `inactivity alerts`...

https://documentation.solarwinds.com/en/success_center/papertrail/content/kb/how-it-works/alerts.htm?cshid=pt-how-it-works-alerts  


> The above instructions cover the setup for alerting on matching events, but what about when no events were sent because something didn't happen? When a cron job, backup, or other recurring job doesn't run, it's not easy to notice the absence of an expected message. But Papertrail can do the noticing for you.

Happy to look into something else here, but that feels like the simplest solution.

analytics-dashboard/ebextensions/qa/90_iamsync.config
indigobravo: The ID is also used to validate the sudo rules being created.

analytics-dashboard/ebextensions/qa/90_iamsync.config
```diff
@@ -0,0 +1,42 @@
+# Configure Elastic Beanstalk EC2 hosts to use iamsync SSH
+# https://github.com/hypothesis/iamsync
+packages:
+  yum:
+    python36.x86_64: []
+    python36-libs.x86_64: []
+    python36-pip.noarch: []
+    python36-tools.x86_64: []
```
indigobravo, line 8: Install python 3.6 packages

analytics-dashboard/ebextensions/qa/90_iamsync.config
indigobravo: The cron schedule has changed, and is now implemented as a system wide `/etc/cron.d` task.

analytics-dashboard/ebextensions/qa/90_iamsync.config
indigobravo: iamsync has been moved to its own public repo, and the file we need `iamsync.py` is downloaded using `curl`. I have chosen not to pin to a specific commit so we don't have to update this extension everytime we update iamsync. However, this can easily be changed if you think pinning to a specific commit is the better option.

analytics-dashboard/ebextensions/qa/90_iamsync.config
```diff
@@ -0,0 +1,41 @@
...
+
+
+files:
+  "/root/iam.yml":
```
robertknight, line 12: Might be better to name this `iamsync.yml` to more obviously indicate the connection to our iamsync tool, rather than it being say some AWS-specific thing which relates to their IAM service.

analytics-dashboard/ebextensions/qa/90_iamsync.config
```diff
@@ -0,0 +1,41 @@
...
+      SHELL=/bin/bash
+      PATH="venv/bin:/usr/sbin:/bin"
+      5,35 * * * * root sleep ${RANDOM: -1} ; python bin/iamsync.py
+
+
+commands:
+  01_create-bin-dir:
+    command: "mkdir /root/bin"
+  02_download-iamsync:
+    command: "curl https://raw.githubusercontent.com/hypothesis/iamsync/main/iamsync.py -o /root/bin/iamsync.py"
```
robertknight, line 37: This currently always fetches and executes the latest version of iamsync. That's convenient in that if we make small internal changes to iamsync.py as they will be picked up automatically. However this might pose a problem in future if we want to make a backwards incompatible change to iamsync and then gradually upgrade the usage across various projects. I don't know how often we're likely to want to do that.

If the need arises we can pin this to a specific release.

robertknight (approved):
I had a suggestion regarding the name of the config file, and a comment about possibly needing to pin the version of iamsync in future if we find that we need to make a backwards incompatible change to the CLI or some major internal change that needs to be rolled out gradually.

analytics-dashboard/ebextensions/qa/90_iamsync.config
indigobravo: Agreed - If we get to a situation where we are likely to break things we can start to pin the version of iamsync the extension downloads.

analytics-dashboard/ebextensions/qa/90_iamsync.config
indigobravo: I will make that change to the code before deploying the extension. I also think it would be nice to move the config file to a know system location (/etc).

PR #215, opened 2021-06-24
SSH Access and Authorization - iamsync

indigobravo:
Further to the question that was raised on the following PR: https://github.com/hypothesis/deployment/pull/214

>Hmm - it's a bit annoying to have this in base64 form in the repo. Would it be possible to either have the file in this repo but in plain text (diffable) form or alternatively download it from somewhere (eg. an S3 bucket). 

I think the simplest solution to make the iamsync code available in all the contexts we use is to add it to the deployment repo. This would enabled us to easily:

1. Install via ansible roles on static EC2 hosts
2. Install via an ebextension on EB managed EC2 hosts

The other alternative is to create a public repo just to host iamsync.

No need to review the code here. It is a straight copy of: https://github.com/hypothesis/playbook/tree/master/integrations/aws/iam. This is more a question about should we host this in the deployment repo.

### DIFF
--- a/tools/aws/iamsync/README.md
+++ b/tools/aws/iamsync/README.md
@@ -1,0 +1,68 @@
+## iamsync.py
+
+A script that generates and maintains local Linux user accounts using AWS IAM as a source of truth. Users added and removed from eligible IAM user groups will be able to login using SSH once a public ssh key has been added to their AWS account.
+
+#### Parameters
+
+**--config**
+
+Path to configuration file.
+
+**--log**
+
+Path to log file.
+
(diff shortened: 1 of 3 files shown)


indigobravo:
Slack thread where we have been discussing the pros and cons of making iamsync public: https://hypothes-is.slack.com/archives/CR3E3S7K8/p1624545711045000

The general consensus is to make it public in its own repo, and I will take a look at what needs to be done to make that happen.

indigobravo:
Closing this PR. I will open another for hosting this in its own repository.

PR #216, opened 2021-06-25
Refactored bin/jenkins for simpler multi-region use.

indigobravo:
I have refactored the conditional statements to be more efficient,
removed the need to specify a region, and built logic into the script
that groups applications we are looking to deploy to multiple regions so
they can be handled differently to those we are only looking to deploy
to a single region.

### DIFF
--- a/bin/jenkins
+++ b/bin/jenkins
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # The name of the Elastic Beanstalk application to manage
 : ${APP:=}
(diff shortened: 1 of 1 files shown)


indigobravo:
After pondering over the changes proposed here. I am not going to be pushing ahead with this route.

1. Running deploys for multiple apps / environments as a single Jenkins job is going to become problematic
    - The only way to run deploys in parallel is to background the process using BASH. This could end up producing confusing output. Running deploys in series will solve the problem, but it will take twice as log to complete.
 2. There is no way to run against a single region.

There are some improvements that can be made to the conditional logic. The following removes the need for nested `if` statements.
```
  if [ "$TYPE" = "redeploy" ] && [ -n "$APP_DOCKER_VERSION" ]; then
    abort "Do not specify \$APP_DOCKER_VERSION for a redeploy"
  fi

  if [ "$TYPE" = "deploy" ] && [ -z "$APP_DOCKER_VERSION" ]; then
    abort "Cannot deploy unless \$APP_DOCKER_VERSION is specified"
  fi
```

PR #218, opened 2021-06-29
EB definition - Changed DeploymentPolicy for H in Canada

indigobravo:
The DeploymentPolicy for H in Canada has been change to Immutable
from RollingWithAdditionalBatch. This change was made because Jenkins
was unable to sync environment updates.


robertknight:
@indigobravo - We've had problems in the past where Immutable deployments were excruciatingly slow. This will probably be less of an issue in Canada due to the smaller number of instances. If this change gets the deployment working then let's move ahead with it for the moment, however we should keep an eye on h deployment times. Do you have a link to the output from the deployment logs where the deployment failed?

PR #285, opened 2021-07-08
metabase-prod - remove bless

indigobravo:
This commit removes the bless extension in preparation for a platform
upgrade.


indigobravo:
The title of this PR is incorrect. This update targeted bouncer-prod.

PR #321, opened 2021-07-12
Updated eb-release to support platform

indigobravo:
This commit adds support for elastic beanstalk platform extensions, which are used to provide custom updates of the Linux platform. We need this to perform updates to Linux level configuration.

More information on the functionality can be found in the AWS docs: https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/platforms-linux-extend.html

A slack discussion has been started here: https://hypothes-is.slack.com/archives/CR3E3S7K8/p1626078418105400

### DIFF
--- a/bin/eb-release
+++ b/bin/eb-release
@@ -75,6 +75,18 @@ if [ -d "${APP}/ebextensions/${APP_ENV}" ]; then
   cp ${APP}/ebextensions/${APP_ENV}/*.config "$EBEXTENSIONS"
 fi
 
+# Conditionally create platform extension hierarchy. Used to provide
+# custom platform configuration on a per app or global basis.
+PLATFORMEXTENSIONS="${SRCBUNDLEDIR}/.platform"
+if [ -d "common/platform" ]; then
+  mkdir -p "$PLATFORMEXTENSIONS"
+  cp -r common/platform/ "$PLATFORMEXTENSIONS"
+fi
+if [ -d "${APP}/platform/${APP_ENV}" ]; then
+  mkdir -p "$PLATFORMEXTENSIONS"
+  cp -r ${APP}/platform/${APP_ENV}/ "$PLATFORMEXTENSIONS"
+fi
+
 # And clean up when we're done...
 trap 'rm -r "$SRCBUNDLEDIR"' EXIT
 


bin/eb-release
```diff
@@ -75,6 +75,18 @@ if [ -d "${APP}/ebextensions/${APP_ENV}" ]; then
   cp ${APP}/ebextensions/${APP_ENV}/*.config "$EBEXTENSIONS"
 fi
 
+# Conditionally create platform extension hierarchy. Used to provide
+# custom platform configuration on a per app or global basis.
+PLATFORMEXTENSIONS="${SRCBUNDLEDIR}/.platform"
```
indigobravo, line 80: Define the `PLATFORMEXTENSIONS` variable which maps to the `.platform` path inside the $SRCBUNDLEDIR

bin/eb-release
```diff
@@ -75,6 +75,18 @@ if [ -d "${APP}/ebextensions/${APP_ENV}" ]; then
...
+if [ -d "common/platform" ]; then
+  mkdir -p "$PLATFORMEXTENSIONS"
+  cp -r common/platform/ "$PLATFORMEXTENSIONS"
+fi
```
indigobravo, line 84: Conditional statement that looks for a `platform` directory inside [common](https://github.com/hypothesis/deployment/tree/master/common). If found `$PLATFORMEXTENSIONS` directory will be created, and the contents of `common/platform/` copied over.

We are dealing with directories in the context of `platform` customisations rather than files that need to be copied in the context of a `ebextension`. Here is an example directory layout we will need to cater for:

```
~/app_source_bundle/
|-- .ebextension
|-- .platform
|    -- nginx
|        -- nginx.conf
```

bin/eb-release
```diff
@@ -75,6 +75,18 @@ if [ -d "${APP}/ebextensions/${APP_ENV}" ]; then
...
+if [ -d "${APP}/platform/${APP_ENV}" ]; then
+  mkdir -p "$PLATFORMEXTENSIONS"
+  cp -r ${APP}/platform/${APP_ENV}/ "$PLATFORMEXTENSIONS"
+fi
```
indigobravo, line 88: Similar to the above, but this section deals with per application environment customisations.

PR #339, opened 2021-07-13
Created common elastic beanstalk platform extensions

indigobravo:
## Overview
We have the opportunity to consolidate a lot of what we are doing with `ebextensions` with elastic beanstalk [platform extensions](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/platforms-linux-extend.html). Platform extensions are superior to ebextensions because they run executable code, and can perform conditional expressions for example.

They are set up in a slightly different way to support differing stages of the elastic beanstalk deployment lifecycle. The following standardised directory layout is required. 

```
.platform/
  -- hooks/
    -- prebuild/
    -- predeploy/
    -- postdeploy/
``` 

This is nice because it gives us the ability to set things up at different stages of deployment. One of the problems I encountered recently was getting papertrail to start before iamsync runs so the logs of the initial run are pushed. Unfortunately, there is no nice way to do this with an ebextension without introducing a synthetic delay.

### Common verses application environment level install

As per ebextensions we have the ability to install an extension at either a common or a application environment level. Installing an extension at an application environment level will take precedence over installing a common level extension.  

## Details
This PR sets out to replace 3 ebextensions that have either been duplicated to support multiple environments, or have been installed in our `common` location. These are:

|Name|Purpose|Install Location|Notes|
|:-|:-|:-|:-|
|100_newrelic-infra.config|Install and configure newrelic infra agent on all elastic beanstalk environments|common||
|02_iamsync.config|Install and configure iamsync ssh|app|Installed to multiple environments, but was always going to be moved to common once installed everywhere|
|50_iptables.config|Creates iptables firewall runs for via proxy services|app|Installed to multiple via / viahtml environments|

The ebextensions listed above will be replaced by the following shell scripts implemented as platform extensions...

**01_newrelic_setup.sh**

A shell script that creates a configuration file, sets up the newrelic repository and installs the infra agent.

**02_via_firewall.sh**

A shell script that creates iptables firewall rules on via and viahtml proxy services. A conditional statement is being used to restrict the environments where this code is executed. Installing this at the common level removes 8 application environment level ebextensions.

**01_iamsync_setup.sh**

A shell script that creates a configuration file, then downloads and executes iamsync. Installed to `predeploy` to enabled other services to start before we run iamsync.


common/platform/hooks/prebuild/02_via_firewall.sh
```diff
@@ -0,0 +1,45 @@
...
+# - Rejecting requests that attempt to access to either RFC1918 private
+#   address space, or the aws metadata service.
+# - Accepting requests specifically to and from the Docker containers
+#   running the service.
+set -eu
+
+# Grab the environment_name value
+ENV_NAME="$(/opt/elasticbeanstalk/bin/get-config container -k environment_name)"
+# Spilt on '-' and extract the second to last record
+APP="$(echo $ENV_NAME | awk -F- '{print $(NF-1)}')"
```
indigobravo, line 23: The name of our via environments are:
1. via-prod
2. via-qa
3. viahtml-prod
4. viahtml-qa
5. lms-via-prod
6. lms-via-qa
7. lms-viahtml-prod
8. lms-viahtml-qa

The awk expression runs a split on the `-` and then returns the second to last record. Given our naming convention that will always result in `via` or `viahtml`.

common/platform/hooks/prebuild/02_via_firewall.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+#
+# 02_via_firewall.sh
+#
+# Script to set iptables firewall runs for via and viahtml applications
```
robertknight, line 5:
```suggestion
# Script to set iptables firewall rules for via and viahtml applications
```

common/platform/hooks/prebuild/02_via_firewall.sh
```diff
@@ -0,0 +1,45 @@
...
+# - Accepting requests specifically to and from the Docker containers
+#   running the service.
+set -eu
+
+# Grab the environment_name value
+ENV_NAME="$(/opt/elasticbeanstalk/bin/get-config container -k environment_name)"
+# Spilt on '-' and extract the second to last record
+APP="$(echo $ENV_NAME | awk -F- '{print $(NF-1)}')"
+
+if [[ $APP = 'via' ]] || [[ $APP = 'viahtml' ]] ; then
```
robertknight, line 25: I wonder if we could apply these rules more broadly in future. How many Docker containers actually need to access EC2 metadata or private IPs? I guess those apps which depend on services within the VPC will? (eg. connections to RDS and Elasticsearch).

common/platform/hooks/prebuild/02_via_firewall.sh
```diff
@@ -0,0 +1,45 @@
...
+  iptables --insert DOCKER-USER --destination 10.0.0.0/8 --jump REJECT
+  iptables --insert DOCKER-USER --destination 172.16.0.0/12 --jump REJECT
+  iptables --insert DOCKER-USER --destination 192.168.0.0/16 --jump REJECT
+  iptables --insert DOCKER-USER --protocol tcp --destination $container_ip_one --jump ACCEPT
+  iptables --insert DOCKER-USER --protocol tcp --source $container_ip_one --jump ACCEPT
+  iptables --insert DOCKER-USER --protocol tcp --destination $container_ip_two --jump ACCEPT
+  iptables --insert DOCKER-USER --protocol tcp --source $container_ip_two --jump ACCEPT
+  iptables --insert DOCKER-USER --protocol udp --destination-port 53 --jump ACCEPT
+  iptables --insert DOCKER-USER --protocol udp --source-port 53 --jump ACCEPT
+  iptables --insert DOCKER-USER --destination 169.254.169.254 --jump REJECT
```
robertknight, line 42: Is this IP the EC2 metadata endpoint?

robertknight (approved):
Looks good in principle. I'm glad to see much less duplication in the repository.

common/platform/hooks/prebuild/02_via_firewall.sh
indigobravo: Thanks you!

common/platform/hooks/prebuild/02_via_firewall.sh
indigobravo: > I wonder if we could apply these rules more broadly in future

I suspect with a little update. These could be applied to all environments, and could potentially improve our security posture, but I am not sure how we could test that.

> I guess those apps which depend on services within the VPC will? (eg. connections to RDS and Elasticsearch).

I have not looking into this, but I would suspect AWS will provide IP ranges for their services that we could choose to allow access to.

Saying that, RDS and Elasticsearch are deployed within our VPCs IP space, so it might just be a case of adding a few IPs and ports to the rules in the context of H and LMS.

_I would say this is a lot more important with the Via because it gives the user a much wider scope to exploit our services._

common/platform/hooks/prebuild/02_via_firewall.sh
indigobravo: Correct. 169.254.169.254 is the AWS [metadata service](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html). Attempts by the Docker container to connect are being rejected.

PR #355, opened 2021-08-10
Metabase prod - Changed instance type - t3.medium

indigobravo:
This commit changes the EC2 instance type being used by the metabase
production environment. The system was previously running t3.small and
was running low on memory.

The system has also crashed a couple of times in the last month or so
because of resource starvation.

An additional 2GB of RAM will be allocated by this change. Taking the
memory available to Metabase from 2GB to 4GB.


indigobravo:
https://hypothes-is.slack.com/archives/C0LUWQQJJ/p1628560800000500

PR #356, opened 2021-08-11
EB Platform Hook - Papertrail logging

indigobravo:
Two updates are provided by this branch...

1. A common platform hook `00_papertrail.sh`. It sets up remote_syslog2 to push a standard set of logs from the hypothesis eb environments to papertrail. These log files are:
    - /var/log/iamsync.log
    - /var/log/nginx/*.log
 2. It also removes the old application specific ebextensions that were once setting up papertrail.

In addition, a number of filters have been created in papertrail to exclude noisy nginx access logs for:
  - via-prod
  - checkmate-prod
  - lms-via-prod
  - lms-viahtml-prod
  - viahtml-prod
  - h-prod
 
Screenshot taken from: https://papertrailapp.com/destinations/13521581/filter
 
Screenshot 2021-08-11 at 14 00 00

### Additional info

The original application specific ebextensions could be used to specify a custom set of logs to be pushed to papertrail. However, similar behaviour can be configured with papertrail filters. There was also a lot of duplication across our deployment repository because of this, and it potentially provides a level of customisation we do not require.

If we decide application specific overrides are needed for papertrail it can be implemented, but I would prefer to see if the `common` level solution defined in this PR works first.

Link to a slack chat where this point has been discussed: https://hypothes-is.slack.com/archives/CR3E3S7K8/p1628673373023600


common/platform/hooks/prebuild/00_papertrail.sh
```diff
@@ -0,0 +1,74 @@
...
+
+remote_archive="https://github.com/papertrail/remote_syslog2/releases/download/v0.20/remote_syslog_linux_amd64.tar.gz"
+local_archive="/usr/local/share/remote_syslog_linux_amd64.tar.gz"
+expected_md5='40d974d4a937868b0abc7e986b5d785c'
+remote_init_file="https://raw.githubusercontent.com/papertrail/remote_syslog2/48afeaaac26d05481fb89ff53dd5522ed25b8c46/examples/remote_syslog.init.d"
+eb_environment_name="$(get-config container -k environment_name || true)"
+aws_metadata_url="http://169.254.169.254/latest/meta-data/instance-id"
+eb_instance_id="$(wget -q -O - $aws_metadata_url)"
+eb_environment_tier=$(echo $eb_environment_name |awk -F- '{print $NF}')
+eb_hostname=${eb_environment_name}_${eb_instance_id}
```
indigobravo, line 30: Defines variables used in the script.

Some of the lines here are long and a bit ugly. I am referring to the `github` URLs more than anything else. If you feel they should be broken down, let me know, and I will provide an update.

common/platform/hooks/prebuild/00_papertrail.sh
```diff
@@ -0,0 +1,74 @@
...
+# It should be noted that nginx logs can generate a lot of data. If
+# needed a filter can be created in papertrail to exclude them.
+#
+# https://documentation.solarwinds.com/en/success_center/papertrail/content/kb/how-it-works/log-filtering.htm
+#
+set -eu
+
+export PATH="$PATH:/opt/elasticbeanstalk/bin"
+
+remote_archive="https://github.com/papertrail/remote_syslog2/releases/download/v0.20/remote_syslog_linux_amd64.tar.gz"
```
indigobravo, line 22: Variable `remote_archive` references a specific release of remote_syslog2 (v0.20) that will not change unless we decide to.

common/platform/hooks/prebuild/00_papertrail.sh
```diff
@@ -0,0 +1,74 @@
...
+local_archive="/usr/local/share/remote_syslog_linux_amd64.tar.gz"
+expected_md5='40d974d4a937868b0abc7e986b5d785c'
```
indigobravo, line 24: The `md5sum` of the remote_syslog2 archive. Used later on in a conditional to check if we have downloaded the correct thing.

common/platform/hooks/prebuild/00_papertrail.sh
indigobravo, line 30: Here we are making calls to elastic beanstalk to get the name of the environment. The AWS metadata service is also being called to obtain the instance id of the EC2 host where the code is being run.

All of this is being used to build the `hostname` and TCP port in the `/etc/log_files.yml` config file.

common/platform/hooks/prebuild/00_papertrail.sh
```diff
@@ -0,0 +1,74 @@
...
+eb_environment_tier=$(echo $eb_environment_name |awk -F- '{print $NF}')
+eb_hostname=${eb_environment_name}_${eb_instance_id}
+
+if [[ $eb_environment_tier = 'prod' ]] ; then
+  papertrail_port="33163"
+fi
+
+if [[ $eb_environment_tier = 'qa' ]] ; then
+  papertrail_port="37197"
+fi
```
indigobravo, line 38: Conditional statement to define the TCP port that will be used to connect to the destination papertrail service. We use two of these. One for `prod` and another for `qa`.

common/platform/hooks/prebuild/00_papertrail.sh
```diff
@@ -0,0 +1,74 @@
...
+
+tar --extract \
+    --file $local_archive \
+    --strip-components 1 \
+    --directory /usr/local/bin \
+    remote_syslog/remote_syslog
+
+
+curl $remote_init_file -Lo /etc/init.d/remote_syslog
+chmod +x /etc/init.d/remote_syslog
```
indigobravo, line 57: Download the init file and set execute permissions.

common/platform/hooks/prebuild/00_papertrail.sh
```diff
@@ -0,0 +1,74 @@
...
+files:
+  - /var/log/eb-docker/containers/eb-current-app/*.log
+  - /var/log/iamsync.log
+  - /var/log/nginx/*.log
+hostname: $eb_hostname
+destination:
+  host: logs.papertrailapp.com
+  port: $papertrail_port
+  protocol: tls
+CONFIG
```
indigobravo, line 70: Create the config file.

common/platform/hooks/prebuild/00_papertrail.sh
```diff
@@ -0,0 +1,74 @@
...
+
+
+systemctl enable remote_syslog
+systemctl restart remote_syslog
```
indigobravo, line 74: Enable and restart the remove_syslog service using systemctl.

The `restart` rather than a `start` is required in our qa environments where deployments are completed on the same EC2 host.

common/platform/hooks/prebuild/00_papertrail.sh
robertknight: As long as we're using remote_syslog2, pinning the version seems wise. A downside is that if they release any security updates we won't get notified.

common/platform/hooks/prebuild/00_papertrail.sh
robertknight: To check my understanding - Papertrail allows submissions on multiple ports and the receiving port is used to tag a log message to indicate which environment it comes from?

robertknight (approved):
LGTM. The `00_papertrail.sh` script looks nice and straightforward. Let's see how we get on with Papertrail-based log filtering.

common/platform/hooks/prebuild/00_papertrail.sh
indigobravo: Pretty much. In total we have 3 `Log Destinations` provided on separate ports. It enables us to have differing settings and filters. They get used for:
- Prod
- QA
- Static Infra

indigobravo:
Thanks Rob!

> LGTM.

I will be deploying Monday(Aug 16th) as I will be concentrating on other tasks today. I want to roll this out slowly and test to make sure we don't lose any log messages.

PR #366, opened 2021-08-31
LMS-Via - Persistent storage

indigobravo:
This is the first cut of a custom config that will present additional
persistent storage to the Docker containers running the LMS-Via service.

The idea is to create a directory on the host and populate it with data
stored in an S3 bucket. The directory then gets mounted to the Docker
container at runtime.


lms-via/platform/qa/hooks/prebuild/10_via_storage.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+#
+# 10_via_storage.sh
+#
+# Script to manage external storage used by lms-via.
+#
+set -eu
+
+mkdir /viastore
```
indigobravo, line 9: Create a directory on the host EC2 server that will be used to store data downloaded from an S3 bucket.

The root filesystem on an EB EC2 host is around `8GB`. In normal operation around `3GB` is used by the operating system and associated services running the Docker container. 

We are only looking to store around `100MB` of data on the EC2 server so I don't think we will hit a problem unless requirements change.

There is the possibility to have the directory backed by another EBS volume. Setting this up would solve the issue of the disk potentially filling up with data downloaded from S3. However, the cost in time to create, format and mount the volume will add to our total deployment time, and I am not sure it is worth it at this stage.

lms-via/platform/qa/hooks/prebuild/10_via_storage.sh
```diff
@@ -0,0 +1,11 @@
...
+
+aws s3 cp s3://viastore/GDrive-Resource-Keys /viastore/
```
indigobravo, line 11: Download file from S3 bucket and save locally.

I am working on the requirements to present `100MB` of storage to the container. As part of my testing, I downloaded a `200MB` file from S3, which took around `3s` to complete.

```
[ianburden@ip-10-1-3-134 ~]$ time aws s3 cp s3://viastore/GDrive-Resource-Keys .
download: s3://viastore/200_meg to ./200_meg

real	0m2.899s
user	0m1.898s
sys	0m1.065s
```

lms-via/Dockerrun.aws.json
```diff
@@ -7,5 +7,11 @@
     {
       "ContainerPort": "9083"
     }
+  ],
+  "Volumes": [
+    {
+      "HostDirectory": "/viastore",
+      "ContainerDirectory": "/viastore"
+    }
   ]
```
indigobravo, line 16: Tell Docker to mount the `/viastore` directory on the EC2 host to `/viastore` on the container.

lms-via/platform/qa/hooks/prebuild/10_via_storage.sh
indigobravo: >Not sure about the bucket and directory names. Maybe change slightly to  via-data so it’s obvious that via is an app name?

Directory name has been changed on Rob's request to: `via-data`

indigobravo:
https://hypothes-is.slack.com/archives/CR3E3S7K8/p1630400247018700

PR #370, opened 2021-09-06
Public Via - Storage Configuration

indigobravo:
Here we are setting up some additional storage for the Public Via
service.

1. 10_via-data.sh
Setup the `via-data.sh` platform hook which downloads data from S3 and
makes it available to the EC2 host running the Docker service.

2. Dockerrun.aws.json
Updated to mount the `/via-data` directory on the EC2 host to the Docker
container running the service.


indigobravo:
We have already been through a similar review for the LMS service: https://github.com/hypothesis/deployment/pull/366 We are basically pushing the same code to the Public service so I am not pushing this through the review machine again.

PR #374, opened 2021-09-16
Papertrail platform hook update

indigobravo:
This commit adds the concept of a region to the Papertrail platform
hook. The region is identified by calling to the AWS metadata service,
and is used when setting the hostname seen by Papertrail.

If a host resides **outside** of our default US region. The hostname will
have a region code included.

If a host resides **inside** our default US region. The hostname will
**not** have a region code included.

The additional information can be used to created groups in Papertrail
that include application logs by a specific region.

---

This update addresses a problem Sean found where logs from the Canada application servers could not be distinguished from those running in the US. [Link to Slack](https://hypothes-is.slack.com/archives/CR3E3S7K8/p1631720070007600)


common/platform/hooks/prebuild/00_papertrail.sh
```diff
@@ -23,17 +23,27 @@ remote_archive="https://github.com/papertrail/remote_syslog2/releases/download/v
...
+
+if [[ $eb_region = 'us-west-1' ]] ; then
+  eb_hostname=${eb_env_name}_${eb_instance_id}
+fi
+
+if [[ $eb_region = 'ca-central-1' ]] ; then
+  eb_hostname=${eb_env_name}-ca_${eb_instance_id}
+fi
+
+if [[ $eb_env_tier = 'prod' ]] ; then
```
indigobravo, line 42: 1. Shortened variable names
2. Standardised on using curl instead of wget for HTTP calls
3. Added a means to identify the region where the script is being run with the `eb_region` variable. This information is then used to add a two letter region code to the hostname. In this case `ca` for our Canada region.

common/platform/hooks/prebuild/00_papertrail.sh
```diff
@@ -23,17 +23,27 @@ remote_archive="https://github.com/papertrail/remote_syslog2/releases/download/v
...
 expected_md5='40d974d4a937868b0abc7e986b5d785c'
 remote_init_file="https://raw.githubusercontent.com/papertrail/remote_syslog2/48afeaaac26d05481fb89ff53dd5522ed25b8c46/examples/remote_syslog.init.d"
-eb_environment_name="$(get-config container -k environment_name || true)"
-aws_metadata_url="http://169.254.169.254/latest/meta-data/instance-id"
-eb_instance_id="$(wget -q -O - $aws_metadata_url)"
-eb_environment_tier=$(echo $eb_environment_name |awk -F- '{print $NF}')
-eb_hostname=${eb_environment_name}_${eb_instance_id}
 
-if [[ $eb_environment_tier = 'prod' ]] ; then
+md_srv="http://169.254.169.254"
```
robertknight, line 27: I assume `md` is short for metadata here. I'd suggest to spell it out explicitly.

common/platform/hooks/prebuild/00_papertrail.sh
```diff
@@ -23,17 +23,27 @@ remote_archive="https://github.com/papertrail/remote_syslog2/releases/download/v
...
+md_doc="$(curl $md_srv/latest/dynamic/instance-identity/document)"
+eb_region="$(echo $md_doc | jq -r .region)"
+eb_instance_id="$(echo $md_doc | jq -r .instanceId)"
+eb_env_name="$(get-config container -k environment_name || true)"
```
robertknight, line 31: When do we expect the `|| true` to be used here?

common/platform/hooks/prebuild/00_papertrail.sh
```diff
@@ -23,17 +23,27 @@ remote_archive="https://github.com/papertrail/remote_syslog2/releases/download/v
...
+eb_env_tier=$(echo $eb_env_name |awk -F- '{print $NF}')
+
+if [[ $eb_region = 'us-west-1' ]] ; then
```
robertknight, line 34: Even though many people reading this will know this, I think it would be worth adding an explicit comment that we're leaving out the region name in our main/default region.

robertknight (approved):
Couple of minor suggestions. Otherwise LGTM.

common/platform/hooks/prebuild/00_papertrail.sh
indigobravo: Thanks - I will update.

common/platform/hooks/prebuild/00_papertrail.sh
indigobravo: In this context the `|| true` is unnecessary. If we get to the point of running a platform hook the `environment_name` must have been set.

I will remove it.

common/platform/hooks/prebuild/00_papertrail.sh
indigobravo: Fair point. I will add a comment.

PR #377, opened 2021-09-22
Websocket tuning - Stage two

indigobravo:
After successfully doubling the resources available to the websocket.
This commit takes it to the next level and doubles the values again.

1. The soft and hard nofile limits imposed by the Docker daemon is being
   overridden. The new setting is 16384:16384 / soft:hard.
2. The EC2 nginx worker_connections value has been increase to 8192 from
   16384.


jon-betts (approved):
Looks good. I'll get an `h` PR for the same

PR #379, opened 2021-10-04
Add the standard system for rebuilding requirements

jon-betts:
Some of our Python dependencies are very out of date in this repo. See:

 * https://github.com/hypothesis/private-issues/issues/76

This PR:

 * Does enough to make this repo look like an app to `hdev`
 * Compiles the existing files in place
 * Removes our pinned dependencies and re-compiles

### Testing notes

On the python side:

 * `hdev install-python`
 * `rm requirements/*.txt`
 * `hdev requirements`
 * The requirements should successfully build
 * `hdev run docker`
 * The docker image should be built


pyproject.toml
```diff
@@ -1,8 +1,3 @@
 [tool.hdev]
 project_name = "deployment"
 project_type = "application"
-
-[tool.hdev.run]
-docker.command="git archive --format=tar.gz HEAD | docker build -t hypothesis/deployment:dev -"
-docker.help="Make the app's Docker image"
-
```
jon-betts: If we have to have a Makefile we might as well do this in the same way as other products for now

seanh (approved):

PR #385, opened 2021-11-03
LMS Jenkinsfile-tasks - All region support

indigobravo:
Updated to support running tasks against `all` the AWS regions we support in a single operation.

Part of the work for: https://github.com/hypothesis/playbook/issues/761

_Caveat: This needs testing. I have tested the basics of the logic locally, but need to run this in our live Jenkins to confirm everything is working as expected. I will find a quiet time when people are not using the system to deploy the update._


lms/Jenkinsfile-tasks
```diff
@@ -2,7 +2,7 @@
 
 def environments = ['qa', 'prod'].join('\n')
 
-def regions = ['us-west-1', 'ca-central-1'].join('\n')
+def region_selector = ['all', 'us-west-1', 'ca-central-1'].join('\n')
```
indigobravo, line 5: 1. Change the name of the list we present when a user clicks the `Build with Parameters` option in Jenkins.
2. Added the `all` option as the default value.

lms/Jenkinsfile-tasks
```diff
@@ -24,6 +24,19 @@ def postSlack(state, env, task, region) {
     slackSend color: colors[state], message: messages[state]
 }
 
+def aws_regions = ['us-west-1', 'ca-central-1']
```
indigobravo, line 27: Created a new list that contains the AWS regions being supported.

lms/Jenkinsfile-tasks
```diff
@@ -24,6 +24,19 @@ def postSlack(state, env, task, region) {
...
+def process_all_regions(environment, timeout, cmd, aws_regions_list) {
+    for (int i = 0; i < aws_regions_list.size(); i++) {
+        def aws_region = aws_regions_list[i]
+        process_region(environment, timeout, cmd, aws_region)
+    }
+}
+
+def process_region(environment, timeout, cmd, aws_region) {
+    sh "bin/eb-task-run lms ${environemnt} ${timeout} \"${cmd}\" ${aws_region}"
+}
```
indigobravo, line 38: Created two functions...

1. process_all_regions - Called when the workflow is run with the `all` option. Makes subsequent calls to the `process_region` function.
2. process_region - Executes the [bin/eb-task-run](https://github.com/hypothesis/deployment/blob/master/bin/eb-task-run) command.

lms/Jenkinsfile-tasks
```diff
@@ -39,7 +52,7 @@ pipeline {
                             'be overridden here. The value is in seconds.',
                defaultValue: 'task-default')
         choice(name: 'REGION',
-               choices: regions,
+               choices: region_selector,
```
indigobravo, line 55: Updated because of the change on line 5

lms/Jenkinsfile-tasks
```diff
@@ -79,7 +92,12 @@ pipeline {
                             input(message: "This task might be dangerous. Do you wish to continue?")
                         }
                         postSlack('start', params.ENV, params.TASK, params.REGION)
-                        sh "bin/eb-task-run lms ${params.ENV} ${timeout} \"${cmd}\" ${params.REGION}"
+
+                        if (params.REGION == 'all') {
+                            process_all_regions(params.ENV, timeout, cmd, aws_regions)
+                        } else {
+                            process_region(params.ENV, timeout, cmd, params.REGION)
+                        }
```
indigobravo, line 100: Conditional statement that checks if the `REGION` value supplied by the user is equal to `all`, and if so, run the `process_all_regions` function.

If the value is not equal to `all`. It will run the `process_region` function.

lms/Jenkinsfile-tasks
```diff
@@ -24,6 +24,19 @@ def postSlack(state, env, task, region) {
     slackSend color: colors[state], message: messages[state]
 }
 
+def aws_regions = ['us-west-1', 'ca-central-1']
+
+def process_all_regions(environment, timeout, cmd, aws_regions_list) {
+    for (int i = 0; i < aws_regions_list.size(); i++) {
```
robertknight, line 30: This runs the `eb-task-run` script sequentially for each region. Does the `eb-task-run` execution encompass the whole operation (ie. is the complete operation run sequentially in each region) or does it just kick off a process which completes asynchronously?

robertknight, line 30: I'm not that familiar with Groovy, but I think `for (aws_region in aws_regions_list) { ... }` would work here. See https://groovy-lang.org/semantics.html#_for_in_loop.

lms/Jenkinsfile-tasks
```diff
@@ -24,6 +24,19 @@ def postSlack(state, env, task, region) {
...
+        def aws_region = aws_regions_list[i]
+        process_region(environment, timeout, cmd, aws_region)
+    }
+}
+
+def process_region(environment, timeout, cmd, aws_region) {
+    sh "bin/eb-task-run lms ${environemnt} ${timeout} \"${cmd}\" ${aws_region}"
```
robertknight, line 37:
```suggestion
    sh "bin/eb-task-run lms ${environment} ${timeout} \"${cmd}\" ${aws_region}"
```

lms/Jenkinsfile-tasks
```diff
@@ -24,6 +24,19 @@ def postSlack(state, env, task, region) {
     slackSend color: colors[state], message: messages[state]
 }
 
+def aws_regions = ['us-west-1', 'ca-central-1']
+
+def process_all_regions(environment, timeout, cmd, aws_regions_list) {
```
robertknight, line 29:
```suggestion
def process_regions(environment, timeout, cmd, aws_regions_list) {
```

Since the regions to use are a parameter, it might not be "all the regions" but some subset.

robertknight (commented):
I found a typo and had a small suggestion for consideration. Feel free to merge once those are addressed.

lms/Jenkinsfile-tasks
indigobravo: This will be an asynchronous operation. I am conscious if we try and run updates in parallel the output of the commands will get mixed together.

In recent times the current operation only takes a few second to complete, so I doubt running sequentially is going to cause a big problem.

We can potentially look at running in parallel if timing becomes a issue. However, I think the solution to that will mirror how the deployment workflow operates. For example, we have an `lms-tasks` workflow that makes calls against a `global` tasks workflow, which would trigger separate jobs each region.

Here is the stage view for lms-tasks. It goes back to the mid Aug, and shows the runs taking up to `12 seconds` to complete. I am hoping that means this update should complete in less that `30 seconds`.

Screenshot 2021-11-03 at 10 41 58

indigobravo: Thanks - Let me take a look.

lms/Jenkinsfile-tasks
indigobravo: Thanks for catching that!

lms/Jenkinsfile-tasks
indigobravo: Sensible suggestion. I will provide an update.

lms/Jenkinsfile-tasks
indigobravo: You were correct. Here is the much easier to read for loop...

```
def process_regions(environment, timeout, cmd, aws_regions_list) {
    for (aws_region in aws_regions_list) {
        process_region(environment, timeout, cmd, aws_region)
    }
}
```

PR #412, opened 2022-05-17
h/Jenkinsfile-tasks updated

indigobravo:
Support for deploying updates to `all` regions has been improved.

1. The default region option has been set to: `all`.
2. The script understands the regions it can deploy to based on
   predefined lists for `prod` and `qa`.
3. The option to deploy to a single specific region is still available.


h/Jenkinsfile-tasks
```diff
@@ -25,11 +25,20 @@ def postSlack(state, env, task, region) {
     slackSend color: colors[state], message: messages[state]
 }
 
-def aws_regions = ['us-west-1', 'ca-central-1']
```
indigobravo, line 28: Removed redundant list.

h/Jenkinsfile-tasks
```diff
@@ -25,11 +25,20 @@ def postSlack(state, env, task, region) {
...
+            process_region(environment, timeout, cmd, it)
+        }
+    }
+
+    if (environment == 'qa') {
+        qa_regions.each {
+            process_region(environment, timeout, cmd, it)
+        }
     }
 }
```
indigobravo, line 43: Amended `process_regions` function.

1. Created new lists that contain the AWS regions we can target for `prod` and `qa` environments.
2. Added conditional statements based on environment. These contain an `each` loop that iterates through either the `prod_regions` or `qa_regions` lists and make subsequent calls to the `process_region` function.

More information on the Groovy `each` loop and the `it` variable can be found here: https://riptutorial.com/groovy/example/32092/each-and-eachwithindex

h/Jenkinsfile-tasks
```diff
@@ -41,9 +50,20 @@ pipeline {
...
+//        We have stopped short of removing all logic relating to the ENV variable
+//        because we want to provide that separation in the future.
+//
+//        The code below provides a default to the ENV variable.
+//        If params.ENV is NOT NULL we use its value.
+//        If params.ENV is NULL we use a value of `prod`.
+//
+//        choice(name: 'ENV',
+//               choices: environments,
+//               description: 'Choose on which environment to run the given task.')
```
indigobravo, line 66: Removed the `choice` parameter called `ENV`. The comment in the code explains the reasoning behind the change.

h/Jenkinsfile-tasks
```diff
@@ -83,6 +103,7 @@ pipeline {
 
                         def cmd = task['cmd']
                         def timeout = task['timeout']
+                        def env = params.ENV :? 'prod'
```
indigobravo, line 106: Defined a new variable called `env` and set it to the value of `params.ENV` if `params.ENV` is NOT NULL. 

If `params.ENV` is NULL a value of `prod` is assigned.

h/Jenkinsfile-tasks
```diff
@@ -91,18 +112,18 @@ pipeline {
...
 
                         if (params.REGION == 'all') {
-                            process_regions(params.ENV, timeout, cmd, aws_regions)
+                            process_regions(env, timeout, cmd)
                         } else {
-                            process_region(params.ENV, timeout, cmd, params.REGION)
+                            process_region(env, timeout, cmd, params.REGION)
                         }
                     }
                 }
```
indigobravo, line 129: Use the new `env` variable instead of `params.ENV`.

h/Jenkinsfile-tasks
```diff
@@ -111,7 +132,7 @@ pipeline {
     }
 
     post {
-        failure { postSlack('error', params.ENV, params.TASK, params.REGION) }
-        success { postSlack('success', params.ENV, params.TASK, params.REGION) }
+        failure { postSlack('error', env, params.TASK, params.REGION) }
+        success { postSlack('success', env, params.TASK, params.REGION) }
```
indigobravo, line 136: Use the new `env` variable instead of `params.ENV`.

robertknight (approved):
Looks good to me. Can you deploy this and try running one of the no-op tasks (eg. get current DB revision) and check that it looks as expected.

indigobravo:
> Looks good to me. Can you deploy this and try running one of the no-op tasks (eg. get current DB revision) and check that it looks as expected.

Thanks Rob. I will deploy the update tomorrow morning before the bulk of the devs come online.

PR #414, opened 2022-05-18
lms/Jenkinsfile-tasks - update

indigobravo:
Support for deploying updates to `all` regions has been improved.

1. The default region option has been set to: `all`.
2. The script understands the regions it can deploy to based on
   predefined lists for `prod` and `qa`.
3. The option to deploy to a single specific region is still available.

In addition, we have decided to remove the `environment` option from the
Jenkins GUI because it is meaningless in the context of our current
deployment and potentially causes confusion.

We have stopped short of removing all logic relating to the ENV
variable because we want to provide that separation in the future.

We now provide a default to the ENV variable:
1. If params.ENV is NOT NULL we use its value
2. If params.ENV is NULL we use a value of `prod`


indigobravo:
This is an extension of https://github.com/hypothesis/deployment/pull/412

https://github.com/hypothesis/playbook/issues/876

PR #418, opened 2022-06-02
LMS / Jenkinsfile-tasks - Revert env option

indigobravo:
This commit reverts the environment option to the Jenkinsfile-tasks
workflow used by LMS.

PR #419, opened 2022-06-02
LMS / Jenkinsfile-tasks - Reverted display name

indigobravo:
As part of a previous commit the environment was incorrectly defaulted
to `prod`. That has now been reverted to enable prod and qa deployments.

This commit reintroduces the environment into the name displayed when
running the job.


indigobravo:
See https://github.com/hypothesis/deployment/pull/418 for more information.

PR #420, opened 2022-08-05
Increase resource types assigned to LMS

indigobravo:
This commit increases the resource types used by our LMS services.

- Production services in both US and CA now use t3.medium instance
  types, which provide 4GB of RAM.

- The QA service in the US now uses t3.small, which provides 2GB RAM.


indigobravo:
https://hypothes-is.slack.com/archives/C0LUWQQJJ/p1659669546316919

PR #429, opened 2022-08-30
Add `make python` and `make requirements`

seanh:
Add `make python` and `make requirements`

Add `make python` and `make requirements` commands similar to those used by the new cookiecutters (https://github.com/hypothesis/cookiecutters).

This means that the `deployment` repo is now self-contained and no longer depends on `tox-pip-sync` and `hdev` to compile its requirements file.

`make requirements` is needed to recompile the `requirements/requirements.txt` file which is used by the `Dockerfile` that we use to run the `deployment` repo's scripts on Jenkins.

`make python` is needed to install the `deployment` repo's version of Python (currently `3.9.7`, as defined by the `.python-version` file and `Dockerfile`). This is needed because `make requirements` needs the right version of Python to be installed.

For good measure I added `make help` and `make clean` as well.

Testing
-------

    make help
    make clean

Install the right version of Python first:

    pyenv uninstall --force 3.9.7
    make python

Re-compiling all the requirements files should work without making any changes to any of the files:

    make --always-make requirements

To upgrade all requirements to the latest versions (this should change the `requirements/*.txt` files):

    make --always-make requirements args=--upgrade

It should be possible to make the `requirements/*.txt` files from nothing but the `requirements/*.in`'s:

    rm -rf .tox requirements/*.txt
    make requirements

Why doesn't it install Python automatically?
--------------------------------------------

This is a pre-existing issue with our implementation of `make requirements` and `make python`, the issue already exists in all our other projects. It's not a big deal: most of the time when you want to run `make requirements` Python will already be installed. If it's not then you just need to run `make python requirements`. In our other projects the other `make` commands (`make dev`, `make test`, etc) aren't affected: those do install Python automatically.

The problem is that `make requirements` should depend on `make python` so that running `make requirements` automatically installs Python. This can't be done currently because `make python` is a `PHONY` target but `make requirements` isn't. `make python` needs to be re-implemented in a non-`PHONY` way.


indigobravo (approved):
Perfect. Thanks!

